PR #33, opened 2024-03-21
Prince/ added temporary static spacing values

prince-deriv:
- added temporary spacing values to be replaced later once figma designs are all connected to `Quill Tokens`


markwylde-deriv (changes requested):
Can we make this really clear it shouldn't be used. Something like `--TEMP_static-spacing-1`, `--NONSTANDARD_static-spacing-1`.

jia-deriv (approved):

prince-deriv:
> Can we make this really clear it shouldn't be used. Something like `--TEMP_static-spacing-1`, `--NONSTANDARD_static-spacing-1`.

sure @markwylde-deriv

meenakshi-deriv (approved):

markwylde-deriv (commented):
How come we are committing the quill.css into the repo. Shouldn't it be considered a "binary" or "build", and therefore be in a `dist` folder and gitignored?

prince-deriv:
> How come we are committing the quill.css into the repo. Shouldn't it be considered a "binary" or "build", and therefore be in a `dist` folder and gitignored?

@markwylde-deriv  technically yes, but we kept it added so that we can check the history of changes, this will allow us to identify if the new token changes/updates will break stuff.

PR #35, opened 2024-03-22
[DTRA] Maryia/DTRA-1034/feat: SegmentedControlSingleChoice

maryia-deriv:
### Usage examples:

desktop_light

mobile_light

### Dark theme:
A **selected segment** will not change its color in dark theme:
    - background color will remain as `--core-color-solid-slate-50`, and text color will remain as `--core-color-solid-slate-1400`.
Only **non-selected segments** will change for dark theme:
    - background color: `--core-color-opacity-white-75`, text color: `--core-color-opacity-white-600`.

desktop_dark

### Keyboard events:
A user will be able to use "Tab" key to enter the focus mode:
mobile_dark
When "Tab" key is pressed, the blue border will appear around a focused option. Next option can be focused when "Tab" is pressed again. After that, a focused non-disabled option can be selected using "Enter" or "Space" key. You can also select options by switching with left/right Arrow keys.

### Animation:
The white selected button is animated when the user selects it by clicking on it. Selecting an option using left/right arrow keys is not animated as per design prototype. Picking a selected option via Story Controls is also not animated.


lib/components/segmented-control/base.tsx
```diff
@@ -0,0 +1,50 @@
...
+            )}
+    </div>
+);
+
+// TODO: need to import placeholder icons from quill-icons
+const Item = ({ className, icon, label, size }: SegmentedControlItemProps) => (
+    <div className={clsx("item", className)}>
+        {icon && <span className="icon">{icon}</span>}
+        {label && (
+            <Typography size={size} as="span">
```
prince-deriv, line 43: let's use the corresponding component here base on what is in the figma. 

`desktop/semantic/typography/body/md/regular/default`

maryia-deriv: <Text /> component, right?

lib/components/segmented-control/segmented-control.scss
```diff
@@ -0,0 +1,81 @@
+@import "../../styles/index.scss";
+
+.segmented-control {
```
farhan-nurzi-deriv, line 3: I think you need to add sliding animation https://www.figma.com/proto/Rgz5XTeUI4HaVWn0P0ACvr/branch/WAP76V7zN1B4byI1NZBrB3/Quill-components?page-id=22504%3A23&type=design&node-id=25738-2505&viewport=694%2C-440%2C0.23&review-number=25&t=hS52nXxZcpqmXGoL-9&scaling=min-zoom&starting-point-node-id=25738%3A2383&show-proto-sidebar=1

lib/components/segmented-control/segmented-control.scss
```diff
@@ -0,0 +1,81 @@
...
+        cursor: pointer;
+
+        &:hover:not(.disabled) {
+            background: var(--core-color-opacity-black-100);
+        }
+        &:active:not(.disabled) {
+            background: var(--core-color-opacity-black-200);
+        }
+        &.disabled {
+            opacity: 0.24;
```
kate-deriv, line 22: Can we use `var(--core-opacity-300)`?

lib/components/segmented-control/segmented-control.scss
maryia-deriv: yeah, i'm working on it, first want to get a preliminary review :)

lib/components/segmented-control/segment.tsx
```diff
@@ -0,0 +1,48 @@
...
+import * as quillIcons from "@deriv/quill-icons";
+import { SegmentedControlProps } from "./base";
+
+interface SegmentProps {
+    className?: string;
+    icon?: string | React.ReactNode;
+    isDisabled?: boolean;
+    isSelected?: boolean;
+    label?: string;
+    onClick?: (event: React.MouseEvent<HTMLDivElement>) => void;
```
kate-deriv, line 13: Are we gonna support onKeyDown? I'm asking, because in my case I have figma design for focused elements (=== tab). That pushed me to thoughts that I should also add support for keyboard events (on 'Enter' or 'Space')

kate-deriv: Please, ignore if it's not relevant

maryia-deriv: for now, we don't any requirements or specification regarding this.

lib/components/segmented-control/base.tsx
```diff
@@ -0,0 +1,47 @@
...
+    className?: string;
+    options: Array<{
+        disabled?: boolean;
+        icon?: string | React.ReactNode;
+        label?: string;
+        selected?: boolean;
+    }>;
+    onChange?: (selectedItemIndex: number) => void;
+    hasContainerWidth?: boolean;
+    size?: Extract<TGenericSizes, "lg" | "md" | "sm">;
```
prince-deriv, line 17: I have added `TRegularSizes` in the generic types, we can use that instead. Thanks

prince-deriv (changes requested):

lib/components/segmented-control/segment.tsx
maryia-deriv: good point, i'll add that👍🏻

lib/components/segmented-control/base.tsx
maryia-deriv: thank you so much :)

lib/components/segmented-control/base.tsx
prince-deriv: Yes @maryia-deriv  `<Text size={md} />`

lib/components/segmented-control/segment.tsx
```diff
@@ -0,0 +1,48 @@
...
+    icon,
+    isDisabled,
+    isSelected,
+    label,
+    onClick,
+    size,
+}: SegmentProps) => {
+    const QuillIcon =
+        typeof icon === "string" &&
+        (quillIcons[icon as keyof typeof quillIcons] as React.ElementType);
```
prince-deriv, line 28: I love the idea of allowing consumers to pass string for icons to leverage on using the Quill Icons, however when might need to check if this will impact the bundle size of our library since we are importing the whole quill icons here.

maryia-deriv: ok, i'll check )

maryia-deriv: good catch, thank you! with this approach, the bundle looks terrifying:
Screenshot 2024-03-27 at 4 30 09 PM
it should be built in 1.30s and not be that huge.
I'll get rid of this approach, which I had to use because of [this known issue with storybook Documentation](https://github.com/storybookjs/storybook/issues/17098), which gets stuck in an infinite loop when you try passing an icon as a component from within a story.
This will be a hard nut to crack.

prince-deriv: There must be a certain configuration in the storybook to avoid that, thanks for checking the bundle size awesome!

lib/components/segmented-control/segmented-control.scss
```diff
@@ -0,0 +1,105 @@
...
+            opacity: var(--core-opacity-300);
+            cursor: not-allowed;
+            pointer-events: none;
+        }
+        &:hover:not(.disabled) {
+            background: color-mix(
+                in srgb,
+                var(--semantic-color-typography-prominent) 8%,
+                transparent
+            );
```
prince-deriv, line 29: Not sure if we really need `color-mix` can you check if this can be achieved using ` --core-color-opacity-black-100`?

maryia-deriv: @prince-deriv in this case, it won't work for dark theme

lib/components/segmented-control/segment.tsx
maryia-deriv: @prince-deriv i made my component support 'placeholder' as 'icon' value :) It won't hurt

lib/components/segmented-control/segment.tsx
```diff
@@ -0,0 +1,67 @@
...
+            <button
+                className={clsx(
+                    "item",
+                    isDisabled && "disabled",
+                    isSelected && "selected",
+                    className,
+                )}
+                onClick={isDisabled ? undefined : onClick}
+                onKeyDown={onKeyDown}
+                ref={ref}
```
kate-deriv, line 56: Why are you avoiding using `disabled` button attribute? 
It's very natural + will help to get rid of checks inside of `onClick` and `onKeyDown`. 
Moreover, you also can use `:disable` in css instead of classname.

kate-deriv: It will be also better for screen-readers, as they will be notified, that button was disabled

lib/components/segmented-control/base.tsx
```diff
@@ -0,0 +1,72 @@
...
+    onChange,
+    hasContainerWidth,
+    size = "md",
+}: SegmentedControlProps) => {
+    const handleKeyboardEvents = (
+        e: React.KeyboardEvent<HTMLButtonElement>,
+        idx: number,
+        ref: RefObject<HTMLButtonElement>,
+    ) => {
+        if (!options[idx].disabled && (e.key === "Enter" || e.key === " ")) {
```
kate-deriv, line 32: what if the button in segment will be disabled?

maryia-deriv: it won't be selected :)

lib/components/segmented-control/segment.tsx
maryia-deriv: @kate-deriv i'm use both CSS and JS for this. We should not get rid of checks inside of onClick and onKeyDown because CSS is something that any user can alter in his Browser Dev Tools. CSS is for beauty (pointer-events) while JS should always be handling the logic (which events are allowed/disabled).

maryia-deriv: good point :) that's good i changed `div` to `button` and able to use it disabled attribute now👍🏻

tsconfig.json
```diff
@@ -9,7 +9,7 @@
           },
         "target": "ES2020",
         "useDefineForClassFields": true,
-        "lib": ["ES2020", "DOM", "DOM.Iterable"],
+        "lib": ["ES2020", "ES2023.Array", "DOM", "DOM.Iterable"],
```
maryia-deriv, line 12: Unlike es2020, es2023 supports `findLastIndex()` array method, which I'm using and its working wee but our current TS linter doesn't recognize it without this setting.

lib/components/segmented-control/segment.tsx
```diff
@@ -0,0 +1,97 @@
...
+            onKeyDown,
+            size,
+        },
+        ref,
+    ) => {
+        const [focused, setFocused] = React.useState(false);
+        const Icon =
+            icon === "placeholder"
+                ? placeholder[size as keyof typeof placeholder]
+                : icon;
```
balakrishna-deriv, line 50: I don't see any use case having placeholder icons here.
Can we pass placeholder icons from storybook instead? This way we can reduce the bundle size of this component. :)

lib/components/segmented-control/base.tsx
```diff
@@ -0,0 +1,129 @@
+import React, { KeyboardEvent, RefObject, useEffect, useState } from "react";
+import { TRegularSizes } from "../../types";
+import { Segment } from "./segment";
+import "./segmented-control.scss";
+
+interface Option {
+    disabled?: boolean;
+    icon?: string | React.ReactNode;
+    label?: string;
```
meenakshi-deriv, line 9: `label` can be` string | ReactNode `as some project might have translation for label.

lib/components/segmented-control/base.tsx
```diff
@@ -0,0 +1,129 @@
+import React, { KeyboardEvent, RefObject, useEffect, useState } from "react";
+import { TRegularSizes } from "../../types";
+import { Segment } from "./segment";
+import "./segmented-control.scss";
+
+interface Option {
+    disabled?: boolean;
+    icon?: string | React.ReactNode;
```
meenakshi-deriv, line 8: @maryia-deriv we can pass icon as a quill Icon component` icon ? : QuillIconComponent`

export type QuillIconComponent = React.ForwardRefExoticComponent<
    Omit<QuillSvgProps, "ref">>;
 
and use it as a component like `<Icon iconSize={} className={}/>`

Then we can pass it in story book as `icon : LabelPairedPlaceholderMdRegularIcon`

lib/components/segmented-control/segment.tsx
maryia-deriv: @balakrishna-deriv I had to add support for 'placeholder' because of a known issue with storybook documentation https://github.com/storybookjs/storybook/issues/17098, which gets stuck in an infinite loop when you try passing an icon as a component from within a story.

lib/components/segmented-control/base.tsx
maryia-deriv: @balakrishna-deriv I had to add support for 'placeholder' because of a known issue with storybook documentation https://github.com/storybookjs/storybook/issues/17098, which gets stuck in an infinite loop when you try passing an icon as a component from within a story.

lib/components/segmented-control/base.tsx
maryia-deriv: thank you) changed to `ReactNode` as its type includes `string` :)

lib/components/segmented-control/base.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+export interface SegmentedControlProps {
+    className?: string;
+    hasAnimation?: boolean;
+    options: Array<Option>;
+    onChange?: (selectedItemIndex: number) => void;
+    size?: TRegularSizes;
+}
+
+const KEY = {
```
kate-deriv, line 27: After both PR will be merged (yours and mine) we can move this object somewhere to utils in order to reuse, because most of the components will react with keyboard events

lib/components/segmented-control/segmented-control.scss
```diff
@@ -0,0 +1,154 @@
...
+                z-index: 2;
+            }
+            &:disabled {
+                opacity: var(--core-opacity-300);
+                cursor: not-allowed;
+                pointer-events: none;
+            }
+            &.focused,
+            &.focused::before {
+                // not using default :focus since we are customizing focus state
```
kate-deriv, line 41: You still can use natural pseudo-class, this will help to get rid of all logic in ._tsx_, where you used `useState` `focused`.
But, you need NOT `:focus`, but `:focus-visible` as it will not apply focus your button after click. If you want to customize, you can remove original style by adding `outline: none;` and add your own

lib/components/segmented-control/segmented-control.scss
```diff
@@ -0,0 +1,154 @@
+@import "../../styles/index.scss";
```
kate-deriv, line 1: It should work without @import "../../styles/index.scss";

kate-deriv: I got such suggestion in my PR)

lib/components/segmented-control/segmented-control.scss
maryia-deriv: unfortunately, i can't get rid of useState `focused` because i need to completely customize focus behaviour, not only after click but also for pressing different keys. For example, pressing "Tab" and "Left/Right" arrow keys should have different behavior, even though they both make buttons be focused.

kate-deriv: Sure, you could ignore my comment if it wasn't relevant)

kate-deriv (approved):

prince-deriv (approved):

jia-deriv (approved):

kate-deriv (approved):

meenakshi-deriv (approved):

balakrishna-deriv (approved):

prince-deriv (approved):

kate-deriv (approved):

balakrishna-deriv (approved):

Nuzhy-Deriv (approved):

kate-deriv (approved):

kate-deriv (approved):

prince-deriv (approved):

kate-deriv (approved):

jia-deriv (approved):

lib/components/SegmentedControl/segmented-control.scss
```diff
@@ -0,0 +1,167 @@
...
+        -webkit-tap-highlight-color: transparent;
+
+        .item {
+            display: flex;
+            justify-content: center;
+            align-items: center;
+            cursor: pointer;
+            position: relative;
+            z-index: 2;
+            min-width: 0;
```
maryia-deriv, line 30: needed to prevent `white-space: nowrap` from breaking css grid

kate-deriv (approved):

lib/components/SegmentedControl/segmented-control.scss
```diff
@@ -0,0 +1,167 @@
...
+            height: var(--size-generic-lg);
+            padding: var(--semantic-spacing-general-xs);
+            border-radius: var(--semantic-borderRadius-md);
+
+            .item {
+                height: var(--size-generic-md);
+                padding: var(--temp-static-spacing-5)
+                    var(--semantic-spacing-general-sm)
+                    var(--temp-static-spacing-5)
+                    var(--semantic-spacing-general-sm);
```
prince-deriv, line 118: I will merge this but let's minimize our css as much as possible. Please apply the fix using a new PR. 
```suggestion
```

lib/components/SegmentedControl/segmented-control.scss
```diff
@@ -0,0 +1,167 @@
...
+            height: var(--size-generic-2xl);
+            padding: var(--semantic-spacing-general-sm);
+            border-radius: var(--semantic-borderRadius-lg);
+
+            .item {
+                height: var(--size-generic-lg);
+                padding: var(--semantic-spacing-general-sm)
+                    var(--semantic-spacing-general-md)
+                    var(--semantic-spacing-general-sm)
+                    var(--semantic-spacing-general-md);
```
prince-deriv, line 137: same here 

```suggestion
```

lib/components/SegmentedControl/segmented-control.scss
```diff
@@ -0,0 +1,167 @@
...
+            height: var(--size-generic-4xl);
+            padding: var(--core-spacing-600);
+            border-radius: var(--semantic-borderRadius-xl);
+
+            .item {
+                height: var(--size-generic-xl);
+                padding: var(--temp-static-spacing-9)
+                    var(--semantic-spacing-general-lg)
+                    var(--temp-static-spacing-9)
+                    var(--semantic-spacing-general-lg);
```
prince-deriv, line 152: Also here 
```suggestion
```

PR #38, opened 2024-03-25
[DTRA-1052] / Kate / Selection control (checkbox) component

kate-deriv:
In this PR  a new component Checkbox (part of Selection control) for Quill UI library was created. 

**Structure:**
Checkbox component is built as a `controlled` one, but it also support `ref` usage. Skeleton is based on `<input type='checkbox'/>` and `<label />` with `<Text/>` Quill UI component inside. Native `<input />` is hidden with the usage of CSS, over it there is a quill icon, which is indicating status (`checked`, `unchecked` and `indeterminate`). `<label/>` is connected with `<input/>` by `id` or `name` properties. This is needed to increase accessibility of screen reader users (will read out loud the label, when the user is focused on the element) and users who have difficulty clicking on very small regions (such as checkboxes) - because when a user clicks the text within the `<label>` element, it toggles the input (this increases the hit area). 
Also, component appearance is reflecting user interactions: `default`, `hover`, `active`, `focused`.

**Properties:**
- Default input attributes
- size (fontSize of the `<Text/>` component in `<label>`
- indeterminate (for setting state)
- isLabelPaired (for adding info icon)
- label (ReactNode or string)
- labelClassName (for additional optional customization)
- onChange (callback function, which will be called as a part of onChange and onKeyDown if it was passed)
- wrapperClassName (for additional optional customization)
- ref

Screenshot 2024-03-27 at 9 14 46 AM
Screenshot 2024-03-27 at 9 14 05 AM


lib/components/Selection control/Checkbox/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+import "./checkbox.scss";
+import {
+    StandaloneSquareCheckFillIcon,
+    StandaloneSquareBoldIcon,
+} from "@deriv/quill-icons";
+import { Text } from "../../Typography";
+
+interface CheckboxProps
+    extends Omit<ComponentProps<"input">, "placeholder" | "defaultChecked"> {
+    fontSize?: "sm" | "md" | "lg" | "xl";
```
prince-deriv, line 18:
```suggestion
    size?: "sm" | "md" | "lg" | "xl";
```

Can we minimize it to `size`?

lib/components/Typography/base.tsx
```diff
@@ -7,6 +7,7 @@ export interface TypographyProps extends ComponentProps<THeaders> {
     as?: ElementType | keyof THeaders;
     bold?: boolean;
     italic?: boolean;
+    is_disabled?: boolean;
```
prince-deriv, line 10:
```suggestion
    disabled?: boolean;
```

lib/components/Selection control/Checkbox/index.tsx
kate-deriv: Sure, will do

lib/components/Typography/base.tsx
kate-deriv: Sure, thank you

lib/components/Selection control/Checkbox/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+interface CheckboxProps
+    extends Omit<
+        ComponentProps<"input">,
+        "placeholder" | "defaultChecked" | "size"
+    > {
+    indeterminate?: boolean;
+    isLabelPaired?: boolean;
+    infoIconClassName?: string;
+    size?: "sm" | "md";
```
prince-deriv, line 27: I have added `TMediumSizes` you can use it instead.

prince-deriv (changes requested):

lib/components/Selection control/Checkbox/index.tsx
kate-deriv: Thank you, applied

lib/components/Typography/text/index.tsx
```diff
@@ -24,6 +25,9 @@ export const Text = ({
             className={clsx(
                 `quill-typography__body-text__size--${size}__weight--${weight}__decoration--${decoration}`,
                 "quill-typography__color--default",
+                {
+                    "quill-typography__color--disabled": disabled,
```
maryia-deriv, line 29: can we use opacity in your component styles instead of changing the text color?

lib/components/Selection control/Checkbox/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+                    )}
+                </div>
+                <label htmlFor={rest.id ?? name}>
+                    <Text
+                        size={size}
+                        as="span"
+                        className={clsx(
+                            "quill-checkbox__label",
+                            {
+                                "quill-checkbox__label--disabled": disabled,
```
maryia-deriv, line 167: can we apply opacity to your component instead of changing the text color (and icon color above) for disabled state?

maryia-deriv (changes requested):

lib/components/Selection control/Checkbox/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+                    "quill-checkbox__icon--default",
+                    {
+                        "quill-checkbox__icon--disabled": disabled,
+                    },
+                )}
+            />
+        );
+
+        const infoIcon =
+            size === "sm" ? (
```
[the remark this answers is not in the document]
kate-deriv: There is a difference between Standalone and Paired. 
`<StandaloneCircleInfoRegularIcon size='md'/> `will be 24x24,
`<LabelPairedCircleInfoMdRegularIcon /> `(also `md`) will be 24x16
Because of different width, I can't replace one icon type with another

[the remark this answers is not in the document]
kate-deriv: Same for sm

lib/components/Selection control/Checkbox/index.tsx
kate-deriv: The reason is that currently only -semantic-color exist in both light and dark mode. If I used color, then it would not work for dark theme. 
Should I leave code like this or return back to usage var(--core-color-opacity-black-300) and etc but without dark theme?

lib/components/Selection control/Checkbox/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+            if (e.key === "Enter" || e.key === " ") {
+                if (isIndeterminate) {
+                    setIsIndeterminate(!isIndeterminate);
+                    setIsChecked(false);
+                } else {
+                    setIsChecked(!isChecked);
+                }
+                onChange?.(e);
+            }
+        };
```
balakrishna-deriv, line 95: Looks like most of the code in these two functions is repeated. Can we create a new function to abstract the logic?

lib/components/Selection control/Checkbox/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+                    )}
+                />
+            ) : (
+                <LabelPairedCircleInfoMdRegularIcon
+                    className={clsx(
+                        "quill-checkbox__info-icon",
+                        infoIconClassName,
+                    )}
+                />
+            );
```
balakrishna-deriv, line 127:
```suggestion
     const InfoIcon =
        size === "sm"
          ? LabelPairedCircleInfoSmRegularIcon
          : LabelPairedCircleInfoMdRegularIcon;
      const infoIcon = (
        <InfoIcon className={clsx("quill-checkbox__info-icon", infoIconClassName)} />
      );
```

lib/components/Selection control/Checkbox/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+            disabled = false,
+            indeterminate = false,
+            isLabelPaired = false,
+            infoIconClassName,
+            size = "sm",
+            label,
+            labelClassName,
+            name,
+            onChange,
+            wrapperClassName,
```
balakrishna-deriv, line 52: can we rename this to className to be consistent with the other components?
```suggestion
            className,
```

lib/components/Selection control/Checkbox/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+    wrapperClassName?: string;
+}
+
+export const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
+    (
+        {
+            checked = false,
+            disabled = false,
+            indeterminate = false,
+            isLabelPaired = false,
```
balakrishna-deriv, line 45:
```suggestion
            hasInfoIcon = false,
```

lib/components/Selection control/Checkbox/index.tsx
kate-deriv: Sure, will abstract, thank you

lib/components/Selection control/Checkbox/index.tsx
kate-deriv: Sure. In that case should I remove other classnames? for label and infoi con?

balakrishna-deriv: className for label and info are fine.

balakrishna-deriv (approved):

lib/components/Selection control/Checkbox/checkbox.scss
```diff
@@ -0,0 +1,109 @@
+@import "../../../styles/index.scss";
```
prince-deriv, line 1: `index.scss` is imported in all css files by default. Can you check if we can remove this?

lib/components/Selection control/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./Checkbox";
```
prince-deriv, line 1: let's change the directory name to `SelectionControl` no spaces

prince-deriv (changes requested):

prince-deriv (approved):

meenakshi-deriv (approved):

prince-deriv (approved):

lib/components/SelectionControl/Checkbox/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+        const [isChecked, setIsChecked] = useState(checked);
+        const [isIndeterminate, setIsIndeterminate] = useState(indeterminate);
+
+        useEffect(() => {
+            setIsChecked(checked);
+        }, [checked]);
+
+        useEffect(() => {
+            setIsIndeterminate(indeterminate);
+        }, [indeterminate]);
```
hasan-deriv, line 66: These useEffects are being triggered based on the state it is changing. It might create infinite loop.

lib/components/SelectionControl/Checkbox/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+        const onChangeHandler = (
+            e:
+                | React.ChangeEvent<HTMLInputElement>
+                | React.KeyboardEvent<HTMLSpanElement>,
+        ) => {
+            if (isIndeterminate) {
+                setIsIndeterminate(!isIndeterminate);
```
hasan-deriv, line 74:
```suggestion
                setIsIndeterminate(prev => !prev);
```

lib/components/SelectionControl/Checkbox/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+                setIsChecked(false);
+            } else {
+                setIsChecked(!isChecked);
```
hasan-deriv, line 77:
```suggestion
                setIsChecked(prev => !prev);
```

lib/components/SelectionControl/Checkbox/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+        const InfoIcon =
+            size === "sm"
+                ? LabelPairedCircleInfoSmRegularIcon
+                : LabelPairedCircleInfoMdRegularIcon;
+
+        const infoIcon = (
+            <InfoIcon
+                className={clsx("quill-checkbox__info-icon", infoIconClassName)}
+            />
+        );
```
hasan-deriv, line 120: can we use this inside the component?

kate-deriv: Sure, will add, thank you

lib/components/SelectionControl/Checkbox/index.tsx
kate-deriv: These `useEffects` are not being triggered based on the state they are changing. They are triggered if the props, which component receive from parent were changed. In my component I'm using `isChecked` from `const [isChecked, setIsChecked] = useState(checked);`

lib/components/SelectionControl/Checkbox/index.tsx
kate-deriv: Thank you, but actually i'm setting new value inside of onChangeHandler, which will be called only onClick/onKeyDown, so I will work with the latest ones.

lib/components/SelectionControl/Checkbox/index.tsx
kate-deriv: Thank you, but actually i'm setting new value inside of onChangeHandler, which will be called only onClick/onKeyDown, so I will work with the latest ones.

meenakshi-deriv (approved):

balakrishna-deriv (approved):

maryia-deriv (approved):

lib/components/SelectionControl/Checkbox/index.tsx
hasan-deriv: It's better to update state in functional way if new state depends on the previous state.

lib/components/SelectionControl/Checkbox/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+interface CheckboxProps
+    extends Omit<
+        ComponentProps<"input">,
+        "placeholder" | "defaultChecked" | "size"
+    > {
+    indeterminate?: boolean;
+    showInfoIcon?: boolean;
+    infoIconClassName?: string;
+    size?: TMediumSizes;
+    label: ReactNode | string;
```
hasan-deriv, line 29:
```suggestion
    label: ReactNode;
```
don't need `string` here. It's already part  of ReactNode

kate-deriv: Thank you, didn't noticed

PR #43, opened 2024-03-29
[QUILL-1543]Nuzhy/labelled, labelless input fields

Nuzhy-Deriv:
Implemented Input Field Component (Labelled & Labelless) in Quill UI.

There are 2 types of Input Text Field.
   - Labelled (when passed `label` prop with inputSize `md`)
   - Labelless (by default)

Properties (all optional):
 - type (HTMLInputTypeAttribute, by default `text`)
 - icon (left side icon)
 - statusIcon (right side icon)
 - inputSize: `md, sm` 
 - status: `neutral, success, error`
 - disabled ( boolean, default: `false`)
 - variant: `outline, fill`, by default `outline`
 - leftStatusMessage (for validation messages)
 - rightStatusMessage (for counters)
 - textAlignment: `left, right` by default `left`
 - label (for Labelled input field with inputSize `md`)

image


lib/components/Input/base/base.scss
```diff
@@ -0,0 +1,349 @@
+@import "../../../styles/quill/spacing.scss";
+@import "../../../styles/quill/font.size.scss";
+@import "../../../styles/quill/color.scss";
+@import "../../../styles/quill/color.semantic.scss";
+@import "../../../styles/quill/color.opacity.scss";
+@import "../../../styles/quill/motion.scss";
```
kate-deriv, line 6: Could you please update imports? As I understand, now they can be written as @quill/...

Nuzhy-Deriv: oh yeah. I didn't commit it. thanks for noticing

lib/components/Input/base/base.scss
```diff
@@ -0,0 +1,350 @@
...
+        align-items: center;
+        padding-block: var(--core-size-50);
+        border-radius: var(--core-borderRadius-400);
+        border-width: var(--borderWidth-xs);
+        font-family: "IBM Plex Sans";
+        overflow: hidden;
+        position: relative;
+        cursor: pointer;
+        border-style: solid;
+        border-width: 1px;
```
kate-deriv, line 79: You have 2 border-width: here and on line 73
+ All border properties(width, style, color) can be written as one with the usage of shorthand

lib/components/Input/base/base.scss
```diff
@@ -0,0 +1,350 @@
...
+
+        &:has(input:disabled) {
```
kate-deriv, line 81: Could you please check if [:has](url) is working in all browsers? I remember in my case it wasn't working in mozilla even it has 92.11% of support in CanIUse

lib/components/Input/base/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+    variant?: "fill" | "outline";
+    leftStatusMessage?: string;
+    rightStatusMessage?: string;
+    textAlignment?: "left" | "center";
+    label?: string;
+    value?: string;
+}
+
+const IconSize = {
+    sm: { width: 24, height: 24 },
```
kate-deriv, line 30: Btw css variables can be used here too

lib/components/Input/base/base.scss
Nuzhy-Deriv: there was no design given for dark theme. but I made sure the colors are not breaking in dark theme.

lib/components/Input/base/base.scss
Nuzhy-Deriv: okay. will check this

lib/components/Input/base/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+    forwardRef,
+    useState,
+} from "react";
+import "./base.scss";
+import React from "react";
+import { TMediumSizes } from "../../../types";
+
+export interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
+    type?: HTMLInputTypeAttribute;
+    icon?: ForwardRefExoticComponent<Omit<QuillSvgProps, "ref">>;
```
kate-deriv, line 16: Will the component be used only with quill icons?

lib/components/Input/text-field/text-field.stories.tsx
```diff
@@ -0,0 +1,160 @@
...
+        statusIcon: StandaloneCircleCheckBoldIcon,
+    },
+};
+export const ErrorMessageTextFieldWithIcons: Story = {
+    args: {
+        placeholder: "Placeholder",
+        variant: "outline",
+        status: "error",
+        leftStatusMessage: "Status message goes here",
+        rightStatusMessage: "0/0",
```
maryia-deriv, line 157: a suggestion for future maintainability: we can create variables for all the repeated strings in the upper scope and pass them to the stories like this :) :
```suggestion
        placeholder, variant, status, leftStatusMessage, rightStatusMessage,
```

lib/components/Input/base/index.tsx
kate-deriv: And why we are rewriting sizes of the icons, can't we just pass token (sm/md) to the icon component?

lib/components/Input/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./base";
```
maryia-deriv, line 1: Shouldn't the directory be called 'input' instead of 'Input'?
*@jia-deriv gave us an example with dir's and files named in kebab-case.

lib/components/Input/base/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+            ...rest
+        },
+        ref,
+    ) => {
+        const [hasValue, setHasValue] = useState(false);
+
+        return (
+            <div className="quill-input__container">
+                <div
+                    data-has-value={hasValue}
```
kate-deriv, line 65: we can use `className` which will be applied if `hasValue` === `true` instead of custom attr:
```
className={clsx(
                    "quill-input__wrapper",
                    {
                        "quill-input__wrapper--has-value": quill-input__wrapper,
                    },
                    className,
                )}
```

lib/components/Input/index.ts
maryia-deriv, line 1: i'm guessing that this file should export both `base` and `text-field`. :)

Also, i thought there should be a `base.tsx` file in the main directory (like in `Typography/`), which would contain the main `Input` component.

maryia-deriv (changes requested):

lib/components/Input/base/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+                    )}
+                >
+                    {Icon && (
+                        <div>
+                            <Icon {...IconSize[inputSize]} />
+                        </div>
+                    )}
+                    <input
+                        {...rest}
+                        type={type}
```
kate-deriv, line 82: Should we restrict some types? What if type='checkbox' will be passed?
I think we should work with type='text'

lib/components/Input/base/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+                            className={clsx(
+                                "label",
+                                `label__status--${status}`,
+                            )}
+                        >
+                            {label}
+                        </label>
+                    )}
+                    {StatusIcon && (
+                        <div>
```
kate-deriv, line 107: Do we need this `<div>`?
Ignore if not relevant

lib/components/Input/base/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+                        ref={ref}
+                    />
+                    {label && inputSize === "md" && (
+                        <label
+                            className={clsx(
+                                "label",
+                                `label__status--${status}`,
+                            )}
+                        >
+                            {label}
```
kate-deriv, line 103: Maybe we can use smth from Typography?

lib/components/Input/base/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+                                {...IconSize[inputSize]}
+                                className={statusIconColors[status]}
+                            />
+                        </div>
+                    )}
+                </div>
+                <div className="message__container">
+                    {leftStatusMessage && (
+                        <p
+                            key={leftStatusMessage}
```
kate-deriv, line 118: Is it better to add `key` attr even outside of array methods?

lib/components/Input/base/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+    icon?: ForwardRefExoticComponent<Omit<QuillSvgProps, "ref">>;
+    statusIcon?: ForwardRefExoticComponent<Omit<QuillSvgProps, "ref">>;
+    inputSize?: TMediumSizes;
+    status?: "neutral" | "success" | "error";
+    disabled?: boolean;
+    variant?: "fill" | "outline";
+    leftStatusMessage?: string;
+    rightStatusMessage?: string;
+    textAlignment?: "left" | "center";
+    label?: string;
```
kate-deriv, line 25: Can label be a component? Or only string

lib/components/Input/text-field/text-field.stories.tsx
```diff
@@ -0,0 +1,160 @@
...
+        layout: "centered",
+    },
+    tags: ["autodocs"],
+    args: {
+        inputSize: "md",
+        status: "neutral",
+        disabled: false,
+        variant: "fill",
+        textAlignment: "left",
+        label: "",
```
kate-deriv, line 21: Do we need empty strings? It's falsy values, will be considered as undefined anyway

lib/components/Input/base/index.tsx
kate-deriv: Please, ignore if not relevant

lib/components/Input/text-field/text-field.stories.tsx
```diff
@@ -0,0 +1,160 @@
...
+    parameters: {
+        layout: "centered",
+    },
+    tags: ["autodocs"],
+    args: {
+        inputSize: "md",
+        status: "neutral",
+        disabled: false,
+        variant: "fill",
+        textAlignment: "left",
```
maryia-deriv, line 20: `inputSize`, `status` and `textAlignment` can be also removed from here because these are default values of the Input component, right? :) Will it work if you remove them?

lib/components/Input/base/index.tsx
maryia-deriv: @Nuzhy-Deriv  i guess `className` should be assigned to the `div` instead of the icon itself in order for the icon to have a correct size and alignment. Otherwise, if it works perfectly as is, `div` looks unnecessary :)

lib/components/Input/base/base.scss
```diff
@@ -0,0 +1,350 @@
...
+            transform: translateY(0);
+        }
+        .peer:focus + .label {
+            transform: translateY(-100%);
+            line-height: var(--core-lineHeight-100);
+            font-size: var(--core-fontSize-50);
+            height: var(--core-size-900);
+            gap: var(--semantic-spacing-gap-sm);
+        }
+        .peer:not(placeholder-shown) + .label {
```
maryia-deriv, line 297:
```
.peer {
    //

    &:placeholder-shown + .label {
        //
    }
    &:focus + .label {
        //
    }
    &:not(placeholder-shown) + .label {
        //
    }
}
```

maryia-deriv (changes requested):

lib/components/Input/base/base.scss
Nuzhy-Deriv: border-width, border-style are not changing but border-color is changing based on status, selectors. that's the reason I'm not writing as one.

lib/components/Input/index.ts
Nuzhy-Deriv: nope. Component directory can be `input`. we are following deriv-ui

lib/components/Input/base/index.tsx
Nuzhy-Deriv: yes, for now I'm allowing `text`, `email`, `password`

lib/components/Input/text-field/text-field.stories.tsx
Nuzhy-Deriv: yes, it will work if we remove. the control won't be available in the storybook

lib/components/Input/base/index.tsx
Nuzhy-Deriv: as per the figma, only string

lib/components/Input/index.ts
jia-deriv: Sorry, my bad we should use pascal case instead of kebab-case if we are following deriv-ui

lib/components/Input/base/base.scss
Nuzhy-Deriv: as I checked, it's working in mozilla

kate-deriv: Thank you)

lib/components/Input/base/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+                        )}
+                        disabled={!!disabled}
+                        onChange={(e) => {
+                            setHasValue(!!e.target.value);
+                            onChange?.(e);
+                        }}
+                        ref={ref}
+                    />
+                    {label && inputSize === "md" && (
+                        <label
```
kate-deriv, line 82: Usually, it's a good practice to bind `label` with `input` by filling in `htmlFor` attribute (it can take `id` or `name` of the `input`). This is needed to increase accessibility of screen reader users (will read out loud the `label`, when the user is focused on the element) and users who have difficulty clicking on very small regions - because when a user clicks the text within the `<label>` element, it toggles the `input` (this increases the hit area).
But it's optional)))

kate-deriv (approved):

lib/components/Input/base/index.tsx
Nuzhy-Deriv: agreed on you @kate-deriv. in the figma given, label is within the input box, plus this component was migrated from Quill-design component which was having `pointer-events: none` for the label. To increase accessibility, will increase the height of label to be same with the input-wrapper and will bind the `input` and `label` with `htmlFor`.

Nuzhy-Deriv: and will remove the `pointer-events: none` from the label.

jia-deriv (approved):

lib/components/Input/base/base.scss
```diff
@@ -0,0 +1,359 @@
...
+@import "@quill/color.scss";
+@import "@quill/color.semantic.scss";
+@import "@quill/color.opacity.scss";
+@import "@quill/motion.scss";
+@import "@quill/font.weight.scss";
+@import "@quill/border.scss";
+@import "@quill/line.height.scss";
+@import "@quill/size.scss";
+
+@keyframes dropIn {
```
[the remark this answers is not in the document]
prince-deriv: We don't have the set guidelines for this yet, But the common naming convention is `camelCase`.

lib/components/Input/text-field/text-field.stories.tsx
jia-deriv: created constant for statuses and variants

lib/components/Input/base/index.tsx
jia-deriv: Changed it to react node

jia-deriv (approved):

prince-deriv (approved):

PR #49, opened 2024-04-01
[DTRA-1041] / Kate / Modal Bottom

kate-deriv:
In this PR a new amazing  component Modal Bottom for Quill UI library was created. This component should be used in MOBILES and TABLETS.
ALERT! ACHTUNG! ВНИМАНИЕ! In all stories I made default view equal to mobile, BUT I wasn't able to do it for documentation tab, so DO NOT CHECK IT THERE, PLEASE HAVE A LOOK TO STORIES IF YOU WANT TEST SMTH.

**Structure:**
This Modal Bottom (Bottom sheet) component will be inserted into the document with the usage of `ReactDOM.createPortal`: component accepts `portalId` (`id` of the modal container) and tries to insert there, but if `portalId` wasn't passed, the modal will try to find any container with `'modal-root' id`. If there is no container with `'modal-root' id` the modal will be insert inside `document.body`. 
Modal consists of two parts: _overlay_ and _modal itself_. _Overlay_ is a `fixed-positioned div`, with `height 100vh`, `width 100vw`, huge `z-index` and `top`, `bottom`, `left`, `right` equal to `zero`, which has a click-listener for closing Modal Bottom.
_Modal itself_, generally, is just a wrapper around passed `children` with a _handlebar_ and _buttons_. _Handlebar_ is optional and has listeners for `swiping up` and `down` (with the usage of `react-swipeable` library). Swipe details can be checked in `swipeConfig` (like delta); also, while user is swiping the modal won't be closed. Swiping up will expand the Modal up to 90% of the screen, swiping down - back to collapsed height. Passed content influence on Modal height, but it can't be more then 90%. If content is too long, then it will be scrollable + swiping down will close the modal, not collapse. 
One button will be rendered by default, second is optional. Modal can be closed by clicking on the secondary button (which is optional) or on the first one if `shouldCloseOnPrimaryButtonClick` is equal to `true`. Also, component accepts optional function, which will be called after user clicks on primary button.
Children can be wrapped with the `<ModalBottom.Header/>` and `<ModalBottom.Body/>` components in order to organize the content inside the modal. Each of them accepts `className` for customization and `<ModalBottom.Image/>` can also be passed `title`, `hasImg` flag, `scr` and `height` properties.

**Properties:**
1. children: Modal's content.
2. isOpened: controls the visibility of the modal.
3. className: className for external tag of the component.
4. showHandleBar: controls the visibility of the handlebar. If handlebar is visible, the modal can be expanded/collapsed by swiping up/down handlebar. Default value is true.
5. showSecondaryButton: controls the visibility of the secondary button.
6. shouldCloseOnPrimaryButtonClick: flag for controlling modal behavior. If it's true, then the modal will be closed after user clicks on the primary button.
7. toggleModal: function for changing state of the visibility of the modal.
8. portalId: ID of the modal container. If it wasn't passed, the modal will try to find any container with 'modal-root' ID. If there is no container with 'modal-root' ID the modal will be insert inside document.body.
9. primaryButtonLabel: label for the primary button.
10. primaryButtonCallback: function which will be called on clicking on primary button.
11. secondaryButtonLabel: label for the secondary button.

Screenshot 2024-04-04 at 3 48 15 PM
Screenshot 2024-04-04 at 3 48 21 PM
Screenshot 2024-04-04 at 3 48 34 PM
Screenshot 2024-04-04 at 3 49 00 PM
Screenshot 2024-04-04 at 3 49 05 PM
Screenshot 2024-04-04 at 3 48 06 PM


lib/components/Modal/Bottom/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+    const animationTimerRef = useRef<ReturnType<typeof setTimeout>>();
+    const swipingTimerRef = useRef<ReturnType<typeof setTimeout>>();
+    const scrollableContainerRef = useRef<HTMLDivElement>(null);
+
+    const modalRoot =
+        (portalId && document.getElementById(portalId)) ||
+        document.getElementById("modal-root") ||
+        document.body;
+
+    const shouldExpand =
```
kate-deriv, line 59: If content is too long, then modal should open on 90% of height. In this case it doesn't make sense to collapse it on swipe down, so, as discussed with designers, it should be closed on swipe down.
This flag is needed to track if modal takes more then 85% of height without expanding

lib/components/Modal/Bottom/modal-bottom.scss
```diff
@@ -0,0 +1,130 @@
...
+            var(--motion-easing-linear);
+
+        &--visible {
+            opacity: var(--core-opacity-1300);
+            transform: translateY(0%);
+            pointer-events: auto;
+        }
+
+        &--expanded {
+            min-height: 90%;
```
kate-deriv, line 49: We can't animate height, so I used well-known css trick for it

lib/components/Modal/Bottom/modal-image.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export const ModalImage = ({
+    children,
+    className,
+    src,
+    height = "var(--temp-static-spacing-210)",
+}: React.PropsWithChildren<ModalBodyProps>) => (
+    <>
+        {src ? (
```
kate-deriv, line 16: I want to give an opportunity to this component to be flexible and to accept ReactNode with img OR to accept just the url, which will be used as a background-image.

lib/components/Modal/Bottom/modal-image.tsx
```diff
@@ -0,0 +1,32 @@
...
+            <div
+                className={clsx("quill-modal-bottom__content-image", className)}
+                style={{
+                    background: `url(${src}) lightgray 50% / cover no-repeat`,
```
kate-deriv, line 20: If src was passed, but it's wrong, then lightgray will be shown

lib/components/Modal/Bottom/modal-image.tsx
```diff
@@ -0,0 +1,33 @@
...
+    height?: string;
+}
+
+export const ModalImage = ({
+    children,
+    className,
+    src,
+    height = "var(--temp-static-spacing-210)",
+}: React.PropsWithChildren<ModalBodyProps>) => (
+    <>
```
maryia-deriv, line 15:
```suggestion
```
no need for this fragment :)

package.json
```diff
@@ -128,6 +128,7 @@
     "dependencies": {
         "@deriv/quill-icons": "^1.19.8",
         "@headlessui/react": "^1.7.18",
-        "usehooks-ts": "^3.0.1"
+        "usehooks-ts": "^3.0.1",
```
maryia-deriv, line 131:
```suggestion
```
i couldn't find 'usehooks-ts' usage anywhere in the PR :)

lib/components/Modal/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./Bottom";
```
maryia-deriv, line 1:
```suggestion
export * from "./bottom";
```
as per @jia-deriv 's suggestion

lib/components/Modal/Bottom/modal-bottom.stories.tsx
```diff
@@ -0,0 +1,476 @@
...
+                    onClick={() => setIsOpen(true)}
+                />
+                <ModalBottom
+                    {...args}
+                    isOpened={isOpen}
+                    toggleModal={setIsOpen}
+                >
+                    <ModalBottom.Title>
+                        {titlePlaceHolderText}
+                    </ModalBottom.Title>
```
maryia-deriv, line 177: Is it possible to unite `ModalBottom.Title` and `ModalBottom.Image` logic under a single `ModalBottom.Header` component which would accept `image` and/or `title` as props?
The reason why I'm suggesting it is that there is not much logic in `ModalBottom.Title` to be worth separating :)

lib/components/Modal/Bottom/modal-bottom.scss
```diff
@@ -0,0 +1,128 @@
...
+        position: absolute;
+        cursor: pointer;
+        top: var(--semantic-spacing-general-sm);
+        left: calc(50% - (var(--core-size-2400) / 2));
+        width: var(--core-size-2400);
+        height: var(--core-size-200);
+        background-color: var(--core-color-opacity-black-300);
+        border-radius: var(--core-borderRadius-100);
+
+        // Pseudo element for increasing swipeable area
```
maryia-deriv, line 64:
```suggestion
        // Pseudo-element for increasing swipeable area
```
nit-pick :)

lib/components/Modal/Bottom/modal-body.tsx
```diff
@@ -0,0 +1,19 @@
...
+    className?: string;
+}
+
+export const ModalBody = ({
+    children,
+    className,
+}: React.PropsWithChildren<ModalBodyProps>) => (
+    <Text
+        size="md"
+        as="div"
```
maryia-deriv, line 14: I suggest using `div` instead of `Text` here and simply let developers pass any elements:  strings, `<Text` components with the `size`, font-weight, etc., - which they like, as `children` :)

maryia-deriv (changes requested):

lib/components/Modal/Bottom/modal-image.tsx
kate-deriv: It's not working without them. There is no syntax error, image is just not rendering)))

package.json
kate-deriv: Yes, correct, I'm not using this package, it was added before me. I just insert comma in order to add react-swipeable library to package.json

lib/components/Modal/index.ts
kate-deriv: Will rename

lib/components/Modal/Bottom/modal-bottom.stories.tsx
kate-deriv: I think you are right, no sense in separating them. Will update PR

lib/components/Modal/Bottom/modal-bottom.scss
kate-deriv: Thank you for noticing)

lib/components/Modal/Bottom/modal-body.tsx
kate-deriv: Actually, I'm not sure about this. I thought content inside ModalBody should have style according to Figma by default.

lib/components/Modal/Bottom/modal-bottom.scss
```diff
@@ -0,0 +1,128 @@
...
+            align-items: center;
+            justify-content: center;
+            flex-direction: column;
+            gap: var(--semantic-spacing-gap-md);
+        }
+    }
+}
+
+//Styles for class which was used in storybook as an example
+.indicator {
```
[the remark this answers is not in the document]
kate-deriv: But the main idea was to show, that component accepts className, which can help to customize. Changing it to style might not be convenient for user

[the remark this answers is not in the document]
kate-deriv: But I'll try to do smth in order to remove styles for storybook from here

lib/components/Modal/Bottom/modal-bottom.stories.tsx
kate-deriv: Oh, okay, we can't unite them) Image and Text should be in separate containers for classnames

lib/components/Modal/Bottom/modal-body.tsx
kate-deriv: Same with ModalTitle

package.json
maryia-deriv: oh)) ok ok)

lib/components/Modal/Bottom/modal-bottom.stories.tsx
kate-deriv: Latest status: unite image and title into the header component with new flags for controlling children

lib/components/Modal/bottom/modal-header.tsx
```diff
@@ -0,0 +1,57 @@
...
+                        {children}
+                    </div>
+                )}
+            </>
+        )}
+        <Heading.H4
+            className={clsx("quill-modal-bottom__content-title", className)}
+        >
+            {title}
+        </Heading.H4>
```
maryia-deriv, line 55:
```suggestion
        {title && <Heading.H4
            className={clsx("quill-modal-bottom__content-title", className)}
        >
            {title}
        </Heading.H4>
        }
```

lib/components/Modal/bottom/modal-header.tsx
```diff
@@ -0,0 +1,57 @@
...
+                    />
+                ) : (
+                    <div
+                        style={style}
+                        className={clsx(
+                            "quill-modal-bottom__content-image",
+                            className,
+                        )}
+                    >
+                        {children}
```
maryia-deriv, line 46:
```suggestion
                        {image}
```
with image prop as `ReactNode`, we can get rid of `children` and `hasImg`: 

```
{src && <div .... }
{image && <div .... }
```
:)

maryia-deriv (changes requested):
please check a couple of last comment, the rest LGTM :)

lib/components/Modal/bottom/modal-header.tsx
kate-deriv: Nice catch, thank you)

lib/components/Modal/bottom/modal-header.tsx
```diff
@@ -0,0 +1,49 @@
+import clsx from "clsx";
+import { Heading } from "../../Typography";
+
+interface ModalBodyProps {
```
ahmadtaimoor-deriv, line 4: Should it be named ModalHeaderProps ?

kate-deriv: Ooops, yes)

ahmadtaimoor-deriv (approved):

lib/components/Modal/bottom/index.tsx
```diff
@@ -0,0 +1,160 @@
+import ReactDOM from "react-dom";
+import { useState, useEffect, useRef } from "react";
+import clsx from "clsx";
+import "./modal-bottom.scss";
+import { useSwipeable } from "react-swipeable";
+import { ModalHeader } from "./modal-header";
+import { ModalBody } from "./modal-body";
+import { Button } from "../../Button";
+
+interface ModalBottomProps {
```
henry-deriv, line 10: interface ModalBottomProps extends HTMLAttributes<HTMLDivElement>

lib/components/Modal/bottom/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+    className,
+    children,
+    showHandleBar = true,
+    showSecondaryButton = false,
+    shouldCloseOnPrimaryButtonClick = false,
+    toggleModal,
+    portalId,
+    primaryButtonLabel,
+    primaryButtonCallback,
+    secondaryButtonLabel,
```
henry-deriv, line 43: {...rest}

maryia-deriv (approved):

lib/components/Modal/bottom/index.tsx
kate-deriv: I think that it's necessary when you are working with buttons, inputs and etc, but div is not so specific. Probably, it might be case-by-case, but I'll apply in order not to restrict future user of the library

lib/components/Modal/bottom/modal-bottom.stories.tsx
```diff
@@ -0,0 +1,474 @@
...
+            description: "Label for the secondary button.",
+            control: { type: "string" },
+        },
+    },
+} satisfies Meta<typeof ModalBottom>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const DefaultModalBottom: Story = {
```
henry-deriv, line 117: im seeing duplicates here. Think its better to define template work off of it so you dont have to replicate same logic in each variant.

https://www.youtube.com/watch?v=FUKpWgRyPlU

starts at minute 11

kate-deriv: Thank you, let me have a look

kate-deriv: Worked perfectly, super, thank you)

henry-deriv (approved):

maryia-deriv (approved):

prince-deriv (approved):

PR #58, opened 2024-04-05
Akmal / feat: add radio button & radio group components

akmal-deriv:
This feature introduces the implementation of radio button and radio group components to enhance user interface options within the project. Radio buttons provide users with a selection mechanism for exclusive choices within a group, while radio groups manage the collective behavior of multiple radio buttons.

Radio Button Component: The radio button component is created to allow users to select a single option from a set of mutually exclusive choices. Each radio button represents a distinct choice, and only one option can be selected at a time.

Radio Group Component: The radio group component encapsulates multiple radio buttons that belong to the same group. It ensures that only one radio button within the group can be selected at any given time, enforcing the mutual exclusivity of choices.

These components offer a user-friendly and intuitive way for users to make selections within forms, surveys, or any other interactive elements where exclusive choices are required. By adding radio button and radio group components, the project enhances its usability and provides a consistent and familiar interface for users to interact with.


lib/components/SelectionControl/radio/radio-button/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+                </span>
+                <span
+                    className={clsx(
+                        "quill-radio-button__label",
+                        classNameLabel,
+                    )}
+                >
+                    {children}
+                </span>
+                {/* TODO: implement info icon component */}
```
ahmadtaimoor-deriv, line 98: Can we remove the TODO

lib/components/SelectionControl/radio/radio-button/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+    classNameLabel,
+    defaultChecked,
+    disabled = false,
+    has_info,
+    id,
+    onChange,
+    value,
+    ...otherProps
+}: React.PropsWithChildren<TRadio>) => {
+    const [checked, setChecked] = React.useState(defaultChecked);
```
ahmadtaimoor-deriv, line 37: Is there a reason to set setChecked with defaultChecked on useEffect as well as initially state setup ?

lib/components/SelectionControl/radio/radio-button/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+                        />
+                    ) : (
+                        <LabelPairedCircleMdRegularIcon
+                            fill={disabled ? "#b8b8b8" : "#7e7e7e"}
+                            data-testid={`dt_unchecked_icon_${value}_${disabled}`}
+                            id={id}
+                        />
+                    )}
+                </span>
+                <span
```
henry-deriv, line 90: i think we use <Text > Component here

lib/components/SelectionControl/radio/radio-button/radio-button.scss
```diff
@@ -0,0 +1,20 @@
+.quill {
+    &-radio-button {
+        display: flex;
+        align-items: center;
+
+        &__label {
+            margin-left: 8px;
+            line-height: 24px;
+            font:
+                16px "IBM Plex Sans",
```
ahmadtaimoor-deriv, line 10: Tokens can be used instead of absolute values

lib/components/SelectionControl/radio/radio-button/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+}: React.PropsWithChildren<TRadio>) => {
+    const [checked, setChecked] = React.useState(defaultChecked);
+
+    React.useEffect(() => {
+        setChecked(defaultChecked);
+    }, [defaultChecked]);
+
+    const onInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+        setChecked(e.target.checked);
+        onChange && onChange(e);
```
henry-deriv, line 45: i think u can just do `onChange?.(e)`

lib/components/SelectionControl/radio/radio-group/radio-group.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+      description: "Hides the third radio button option.",
+    },
+  },
+};
+
+export default RadioGroupMeta;
+
+const defaultArgs: RadioGroupProps = {
+  name: "radioGroup",
+  onToggle: () => console.log("Toggled!"),
```
henry-deriv, line 58: import { fn } from "@storybook/test";
better to do `fn()` gives more information in storybook instead of just a console.log

lib/components/SelectionControl/radio/radio-button/radio-button.scss
```diff
@@ -0,0 +1,20 @@
+.quill {
+    &-radio-button {
+        display: flex;
+        align-items: center;
+
+        &__label {
```
henry-deriv, line 6: should not need this css after switching to Text component

lib/components/SelectionControl/radio/radio-button/radio-button.scss
```diff
@@ -0,0 +1,20 @@
...
+            line-height: 24px;
+            font:
+                16px "IBM Plex Sans",
+                sans-serif;
+        }
+        &__input {
+            display: none;
+        }
+        &__info {
+            margin-left: 8px;
```
henry-deriv, line 17: var(--core-spacing-400)

lib/components/SelectionControl/radio/radio-button/index.tsx
```diff
@@ -0,0 +1,113 @@
+import React from "react";
+import clsx from "clsx";
+import {
+    LabelPairedCircleMdRegularIcon,
+    LabelPairedCircleDotMdFillIcon,
+    LabelPairedCircleInfoMdRegularIcon,
+} from "@deriv/quill-icons";
+import "./radio-button.scss";
+
+type TRadio = {
```
kate-deriv, line 10: In this repo `interface` is used, it's better to stick to it in this case. Also, maybe it's better to extend `input` component props?

lib/components/SelectionControl/radio/radio-button/index.tsx
akmal-deriv: The info icon component is not there yet, I think we should remove it after it's implemented

lib/components/SelectionControl/radio/radio-button/index.tsx
akmal-deriv: The value is changed from the outside of this component, therefore, it needs the useEffect to update it.

lib/components/SelectionControl/radio/radio-button/index.tsx
akmal-deriv: Yeah, it's the same

lib/components/SelectionControl/radio/radio-button/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+    }, [defaultChecked]);
+
+    const onInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+        setChecked(e.target.checked);
+        onChange && onChange(e);
+    };
+
+    const handleIconClick = (e: React.MouseEvent<HTMLSpanElement>) => {
+        if (!disabled) {
+            const synthesizedEvent = new Event("change", { bubbles: true });
```
kate-deriv, line 51: Am I'm right that when user clicks on `span`, `input` should catch `onChange` event? If yes, maybe we can use `<label>` instead of `span`? In this case you can bind them with `name` or `id`, so whenever user clicks on `label`, `input` will be checked. Also, there is another way: you can position `span` just over the `input` and make span non-interactive (`pointer-events: none;`), so actually, user will react with input inself

lib/components/SelectionControl/radio/radio-button/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+                htmlFor={id}
+                className={clsx("quill-radio-button", className, {
+                    "quill-radio-group__item--active": checked,
+                })}
+                data-testid="dt_quill_radio_button"
+            >
+                <input
+                    className="quill-radio-button__input"
+                    type="radio"
+                    onChange={onInputChange}
```
kate-deriv, line 69: Should it work with onKeyDown too? onFocus

lib/components/SelectionControl/radio/radio-group/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+                            >
+                                {item.props.label}
+                            </RadioButton>
+                        </ItemWrapper>
+                    ))}
+        </div>
+    );
+};
+
+const Item = ({
```
maryia-deriv, line 86: perhaps it would be great to move ItemWrapper and Item to a separate file/-s since 3 components, however small, is too much for 1 file IMHO :)

lib/components/SelectionControl/radio/radio-button/radio-button.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+  has_info: boolean;
+  id: string;
+  className: string;
+  classNameLabel: string;
+  classNameInfo: string;
+}
+
+const Template: React.FC<RadioButtonProps> = (args) => <RadioButton {...args}>Option</RadioButton>;
+
+const defaultArgs: RadioButtonProps = {
```
maryia-deriv, line 52: `defaultArgs` can be replaced with `RadioButtonMeta.args` property, so that you don't have to destructure them in every story :)

maryia-deriv (changes requested):

lib/components/SelectionControl/radio/radio-button/index.tsx
akmal-deriv: This is not specifically for the span or label. There's a custom icon for input, whenever it's clicked I trigger the hidden input to trigger logic changes across all radio buttons across the group. The sync between all the buttons is done via the hidden input.

lib/components/SelectionControl/radio/radio-group/index.tsx
akmal-deriv: I think it's fine, these are microscopic component lol

lib/components/SelectionControl/radio/radio-button/index.tsx
akmal-deriv: It shouldn't matter on this input, it is hidden itself, but let me check on the actual button.

kate-deriv: I mean not input itself, but the visual effect, that it can be focused (e.g. around icon)

lib/components/SelectionControl/radio/radio-button/index.tsx
kate-deriv: I mean you can still render `input`, but removing `display: none` and hide by positioning icon over it. If you add `pointer-events: none;` to the icon, user will interact with `input` itself, so in that case there will be no need in extra logic

lib/components/SelectionControl/radio/radio-button/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+    const handleIconClick = (e: React.MouseEvent<HTMLSpanElement>) => {
+        if (!disabled) {
+            const synthesizedEvent = new Event("change", { bubbles: true });
+            const input = e.currentTarget.previousSibling as HTMLInputElement;
+            input.dispatchEvent(synthesizedEvent);
+        }
+    };
+
+    return (
+        <div>
```
kate-deriv, line 58: Do we need this empty div?

lib/components/SelectionControl/radio/radio-button/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+                    value={value}
+                    {...otherProps}
+                />
+
+                <span onClick={handleIconClick}>
+                    {checked ? (
+                        <LabelPairedCircleDotMdFillIcon
+                            fill={disabled ? "#b8b8b8" : "#000000"}
+                            data-testid={`dt_checked_icon_${value}_${disabled}`}
+                            id={id}
```
kate-deriv, line 81: Should `id` be among `input` props? Usually we bind `label` with `input`.
Also, if `id` goes to `input` among `{...otherProps}`, then we'll have duplicated `id` which will cause error

lib/components/SelectionControl/radio/radio-button/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+};
+
+const RadioButton = ({
+    children,
+    className,
+    classNameInfo,
+    classNameLabel,
+    defaultChecked,
+    disabled = false,
+    has_info,
```
kate-deriv, line 32: Should it accept size? In Figma it can be sm and md

lib/components/SelectionControl/radio/radio-button/radio-button.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+  component: RadioButton,
+  parameters: { layout: "centered" },
+  tags: ["autodocs"],
+  argTypes: {
+    id: {
+      description: "Unique identifier for the radio button element.",
+    },
+    defaultChecked: {
+      description: "Sets the initial checked state of the radio button.",
+      control: { type: "boolean" },
```
kate-deriv, line 15: Maybe it's better to add options for some argTypes in order to give the ability to toggle between some values? E.g. 
```
        defaultChecked: {
            description: "Sets the initial checked state of the radio button.",
            options: ["true", "false"],
            control: { type: "boolean" },
        },
```

lib/components/SelectionControl/radio/radio-button/index.tsx
akmal-deriv: Hmm, this sounds too fragile solution to me. If something goes wrong with the layout, the input will be duplicated on the screen which is not a good UX. I might be misunderstanding what you mean though, but I think hiding input is a better option here

lib/components/SelectionControl/radio/radio-button/index.tsx
akmal-deriv: Well, I'd say the devs should also be responsible not to pass duplicated IDs or params where it's not required, I want to keep otherProps for the component flexibility though. But I'll check if there was a need to pass ID here to icons or it can just be passed to input directly.

lib/components/SelectionControl/radio/radio-button/index.tsx
kate-deriv: Okay👍

lib/components/SelectionControl/radio/radio-button/index.tsx
akmal-deriv: Hmm, I'll check. I didn't see small size in figma. I'll look closer and check.

lib/components/SelectionControl/radio/radio-group/__tests__/radio-group.test.tsx
```diff
@@ -0,0 +1,111 @@
+import React from "react";
+import { render, screen, fireEvent } from "@testing-library/react";
```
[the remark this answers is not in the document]
akmal-deriv: For some reason it was impossible to do some tests with userEvent here, so I had to switch to fireEvent to be able to add required checks.

lib/components/SelectionControl/radio/radio-button/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+    classNameLabel?: string;
+    defaultChecked?: boolean;
+    disabled?: boolean;
+    has_info?: boolean;
+    id?: string;
+    name?: string;
+    onChange?: (e: React.ChangeEvent<HTMLInputElement>) => void;
+    required?: boolean;
+    value?: string;
+    size?: "sm" | "md";
```
kate-deriv, line 26: we have type for it:
import { TMediumSizes } from "../../../../types";

lib/components/SelectionControl/radio/radio-button/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+    LabelPairedCircleDotSmFillIcon,
+    LabelPairedCircleInfoMdRegularIcon,
+    LabelPairedCircleInfoSmRegularIcon,
+    LabelPairedCircleMdRegularIcon,
+    LabelPairedCircleSmRegularIcon,
+} from "@deriv/quill-icons";
+import { Text } from "../../../Typography";
+import "./radio-button.scss";
+
+interface IRadio {
```
kate-deriv, line 14: Should it extend `ComponentPrpos<'input'>`?

lib/components/SelectionControl/radio/radio-group/radio-group.stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+    onToggle?: (e: React.ChangeEvent<HTMLInputElement>) => void;
+    selected?: string;
+    shouldWrapItems?: boolean;
+}
+
+const RadioGroupItem = ({
+    value = "",
+    label = "",
+    disabled = false,
+    hidden = false,
```
maryia-deriv, line 18: do you think it will be beneficial to add all these default values to props of `RadioGroup.Item`? :) this way you till not need to have this `RadioGroupItem` wrapper

lib/components/SelectionControl/radio/radio-button/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+            const synthesizedEvent = new Event("change");
+            const input = e.currentTarget.previousSibling as HTMLInputElement;
+            input.dispatchEvent(synthesizedEvent);
+        }
+    };
+
+    const getIcon = () => {
+        if (checked) {
+            return size === "sm" ? (
+                <LabelPairedCircleDotSmFillIcon
```
kate-deriv, line 69: Seems that in Figma designers used StandAlone icons, not LabelPaired for radio btn.
`StandaloneCircleDotFillIconquill` icon component accept `size`, so there will be no need in passing width and height.
Screenshot 2024-04-09 at 12 26 11 PM

lib/components/SelectionControl/radio/radio-button/radio-button.scss
```diff
@@ -0,0 +1,32 @@
+.quill {
+    &-radio-button {
+        display: flex;
+        align-items: center;
+
+        &__label {
+            margin-left: var(--core-spacing-400);
+            cursor: pointer;
+        }
+        &__icon svg{
```
maryia-deriv, line 10:
```suggestion
        &__icon svg {
```

maryia-deriv (changes requested):

lib/components/SelectionControl/radio/radio-button/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+            <Text
+                size={size}
+                as="span"
+                className={clsx("quill-radio-button__label", classNameLabel)}
+            >
+                {children}
+            </Text>
+            {/* TODO: implement info icon component */}
+            {has_info && (
+                size === "sm" ? (
```
kate-deriv, line 139: Can replace with:
 ```
       const InfoIcon =
            size === "sm"
                ? LabelPairedCircleInfoSmRegularIcon
                : LabelPairedCircleInfoMdRegularIcon;
                
                ....
                render (
                ....
                {hasInfo &&  <InfoIcon
                        className={clsx(
                            "quill-radio-button__info",
                            classNameInfo,
                        )}
                        data-testid="dt_quill_radio_button_info"
                    />}
                    ...)
```

File: lib/components/SelectionControl/radio/radio-button/radio-button.stories.tsx [hunk 25]
```diff
@@ -0,0 +1,85 @@
...
+            description: "CSS class name applied to the informational content.",
+        },
+    },
+    description:
+        "The RadioButton component is used to create radio button elements for user selection within a group.",
+};
+
+export default RadioButtonMeta;
+
+interface RadioButtonProps {
```
kate-deriv, line 51: you can use `meta args` or `React.ComponentProps<typeof RadioButton>`

lib/components/SelectionControl/radio/radio-button/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+import { Text } from "../../../Typography";
+import "./radio-button.scss";
+
+interface IRadio {
+    className?: string;
+    classNameInfo?: string;
+    classNameLabel?: string;
+    defaultChecked?: boolean;
+    disabled?: boolean;
+    has_info?: boolean;
```
kate-deriv, line 20: hasInfo

lib/components/SelectionControl/radio/radio-button/index.tsx
kate-deriv: Please, ignore if irrelevant

jia-deriv (approved):

lib/components/SelectionControl/radio/radio-button/index.tsx
akmal-deriv: I don't think there's any need for that. Just makes type gymnastics harder without any benefit.

lib/components/SelectionControl/radio/radio-group/radio-group.stories.tsx
akmal-deriv: This is just here because of typescript gymnastics, some of these values can't be assigned directly to a div element as it's not part of HTMLElement type, so I added the default values here to get rid of typescript complains. I don't think there are any benefits.

lib/components/SelectionControl/radio/radio-button/index.tsx
akmal-deriv: I'm not sure if this is the right way designer used because it's supposed to be an icon with a label, not standalone but since it has a size property, I switched to it. But again, designers shouldn't use standalone icons in a design with label. This defeats the purpose of having separate icons which are visually identical.

lib/components/SelectionControl/radio/radio-button/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+            data-testid="dt_quill_radio_button"
+        >
+            <input
+                className="quill-radio-button__input"
+                type="radio"
+                onChange={onInputChange}
+                checked={checked}
+                disabled={disabled}
+                value={value}
+                {...otherProps}
```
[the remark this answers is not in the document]
akmal-deriv: No

lib/components/SelectionControl/radio/radio-button/index.tsx
akmal-deriv: Why? There are several ways to achieve the same thing in programming. It's not an exact science...

henry-deriv (approved):

lib/components/SelectionControl/radio/radio-button/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+        }
+    };
+
+    const getIcon = () => {
+        if (checked) {
+            return (
+                <StandaloneCircleDotBoldIcon
+                    data-testid={`dt_checked_icon_${value}_${disabled}`}
+                    fill={disabled ? "#b8b8b8" : "#000000"}
+                    iconSize={size}
```
prince-deriv, line 71: Why this icon doesn't have the `tabIndex` and the other one has?

prince-deriv (approved):

lib/components/SelectionControl/radio/radio-button/index.tsx
akmal-deriv: I want to avoid selecting the selected option. Since there's no point to click the option which is already selected, I only added tabIndex to the icon which is used for non-selected options.

PR #60, opened 2024-04-09
[DTRA-1162] / Kate / Selection control (checkbox group)

kate-deriv:
In this PR a new amazing component Checkbox multiple (group) for Quill UI library was created.

**Structure:**
Inside wrapper div nested Checkboxes (single) can be rendered (mapped) based on passed config. Config for building Parents and Children Nodes is actually an array, where each object in array represents Parent Node and includes obligational fields for `id` and `label`. Optional are fields which are named the same way as `props` from Checkbox (single). Also, optional is field `children`, which can include an array with fractal-like Child Nodes objects. Please note, that if you want to change config passed in props, then in order to notify CheckboxGroup component about changed you need to pass config as new one (e.g. with spread operator).
There are 3 key functions inside CheckboxGroup component:

1. `modifyConfig()` is needed in order to update config in case if a) children have field `checked:true` and it's not reflected in the Parent Node config part or 2) parent has `disabled:true`, but children - not.
2. `handleChildChange()` is needed to change state of children (in config copy in state) when user click on them. Also, it's calling `modifyConfig()` inside to update patent.
3. `handleParentClick()` is needed to change state of parent (in config copy in state) when user click on it. Also, it'll update children state based on parent.

**Properties:**
1. `className`: for external tag of the component.
2. `checkboxGroupConfig`: array with objects for mapping children.

Screenshot 2024-04-10 at 3 27 57 PM


lib/components/SelectionControl/checkbox/checkbox-group/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+export const CheckboxGroup = ({
+    className,
+    checkboxGroupConfig,
+}: CheckboxGroupProps) => {
+    const [checkBoxItemsConfig, setCheckBoxItemConfig] =
+        useState(checkboxGroupConfig);
+
+    useEffect(() => {
+        const configCopy = structuredClone(checkBoxItemsConfig);
```
kate-deriv, line 28: It has 93.59% support (CanIUse). Screenshot from mdn web_docs is pinned below. Has polyfill.
In case if it's considered unreliable, I can still add lodash library to projects dependencies and use a cloneDeep() function. Still, the downside is that it requires to add the library to your project, making it an external dependency.
Screenshot 2024-04-11 at 8 06 04 AM

lib/components/SelectionControl/checkbox/checkbox-group/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+                if (item.disabled) {
+                    item.children.forEach(
+                        (subItem) => (subItem.disabled = true),
+                    );
+                }
+            }
+        });
+    };
+
+    const handleChildChange = (
```
ahmadtaimoor-deriv, line 66: Just a thought , 
I see handleChildChange and handleParentChange shares a lot of similar functionality . Do you think we can create a single function for it.

hasan-deriv (approved):

jia-deriv (approved):

lib/components/SelectionControl/checkbox/checkbox-group/index.tsx
```diff
@@ -0,0 +1,145 @@
+import { useState, useEffect } from "react";
+import clsx from "clsx";
+import "./checkbox-group.scss";
+import { Checkbox } from "../checkbox-single";
+import { CheckboxProps } from "../checkbox-single";
```
prince-deriv, line 5: I will merge this PR, but can we apply this change? 

```suggestion
import { Checkbox, CheckboxProps } from "../checkbox-single";
```

PR #71, opened 2024-04-18
Jia/_/ActionSheet controlled fix

jia-deriv:
- **fix: controlled action sheet isOpen state**


lib/components/ActionSheet/root/index.tsx
```diff
@@ -40,6 +40,9 @@ const Root = ({
         if (!isOpen) {
             handleClose();
         }
+        if (isOpen) {
```
prince-deriv, line 43:
```suggestion
        else
```

prince-deriv (approved):

PR #73, opened 2024-04-19
[DTRA] Henry/dtra 1149/snackbars queue

henry-deriv:


lib/components/Snackbar/index.tsx
```diff
@@ -1,99 +1,28 @@
...
-                        </Text>
-                    </div>
-                    {actionText && (
-                        <Button
-                            variant="tertiary"
-                            label={actionText}
-                            color="white"
-                            onClick={handleActionClick}
+        <div className="snackbar--container">
+            {queue.map((item: SnackbarProps, index) => {
```
maryia-deriv, line 10:
```suggestion
            {queue.slice(0, 1).map((item: SnackbarProps, index) => {
```
just a shorter approach allowing to get rid of if condition inside map :)

lib/components/Snackbar/Snackbar.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { Text } from "../Typography";
+import clsx from "clsx";
+import { Button } from "../Button";
+import { LabelPairedXmarkSmBoldIcon } from "@deriv/quill-icons";
+import { useSnackbar } from "@hooks/useSnackbar";
+
+export interface SnackbarProps extends HTMLAttributes<HTMLDivElement> {
+    icon?: ReactNode;
+    id: string;
+    isShown: boolean;
```
prince-deriv, line 11: I would like to understand if this props is acting as a flag to check whether the snackbar has been shown already otherwise, I think we can use `isVisible` instead.

lib/components/Snackbar/snackbar.stories.tsx
```diff
@@ -37,103 +59,165 @@ const meta = {
...
+To add snackbar anywhere in your application, import addSnackbar function from useSnackbar hook and customize your snackbar by providing desired props.
+
+\`\`\`javascript
+import { useSnackbar } from "@hooks/useSnackbar";
+
+<Button 
+    size="lg" 
+    label={"Add Snackbar"}
+    onClick={() => {
+        addSnackbar({
```
maryia-deriv, line 113:
```suggestion
        useSnackbar.addSnackbar({
```

lib/providers/snackbar/snackbarContext.tsx
```diff
@@ -0,0 +1,11 @@
+// SnackbarContext.js
```
prince-deriv, line 1: let's rename the file to `snackbar-content.tsx`

lib/providers/snackbar/snackbarContext.tsx
```diff
@@ -0,0 +1,11 @@
...
+import { SnackbarProps } from "../../components/Snackbar/Snackbar";
```
prince-deriv, line 2: let's use alias when importing

```suggestion
import { SnackbarProps } from "@components/Snackbar/Snackbar";
```

prince-deriv (approved):

lib/components/Snackbar/snackbar.stories.tsx
henry-deriv: ah yeah i forgot to put this in documentation

const { addSnackbar } = useSnackbar();

lib/components/Snackbar/snackbar.stories.tsx
```diff
@@ -37,103 +59,165 @@ const meta = {
...
         docs: {
             story: {
-                height: "100px",
+                height: "200px",
             },
+            page: () => (
+                <>
+                    <Markdown>
+                        {`#How to use snackbar component
+First import SnackbarProvider and SnackbarController in root of your application and wrap SnackbarController inside SnackbarProvider as shown below.
```
maryia-deriv, line 84: It's a very interesting approach with the `useSnackbar` hook and Context provider. :)
I like the idea, however, I doubt that we really need the context or that we need to wrap the entire application in this Context provider just to use Snackbars alone.
In my opinion, the context and the hook are not really necessary to use for managing the state of snackbars in this library.
The state of the Snackbars queue should be managed outside the library - in our application - including removal, adding and other manipulations with snackbars.
So to make stories work, this state can be handled using `useState` inside stories.
`SnackbarController` can be renamed into `SnackbarsQueue` and turned into a regular an exported component that would accept `snackbars` prop as an array of objects for example, and it would notify the parent about the removal of a particular snackbar using a callback.
@balakrishna-deriv @akmal-deriv @prince-deriv  What do you guys think?

maryia-deriv (changes requested):

lib/providers/snackbar/snackbarContext.tsx
henry-deriv: you want to change context and provider files to `snackbar-context.tsx` and `snackbar-provider.tsx`?

lib/components/Snackbar/Snackbar.tsx
henry-deriv: i put "isShown" because provider maintains, manages, and controls whether or not a snackbar in the queue is shown already or not. 

I can keep isShown on provider side and isVIsible on component side.

What do you think?

lib/components/Snackbar/snackbar.stories.tsx
prince-deriv: I agree on you @maryia-deriv, we need to keep the components dumb and simple.

balakrishna-deriv: @maryia-deriv @prince-deriv I asked @henry-deriv to work on this component. Considering that we are splitting the apps into standalone apps like dtrader, trader's hub, p2p, etc.., we need to abstract the logic of queuing in one place. Otherwise, we will have to reimplement the same thing in different repos.

balakrishna-deriv: Snackbar is a dumb component. It can be used separately without any queuing as well.

akmal-deriv: I'd agree that the usage of any component should be as simple as possible, which is not the exact case here, however, given the point that we are indeed moving towards multiple repos approach that are still visually a part of a single application, we will have to manage the queue in a single place, not inside the multiple apps, and I guess this is one of the better approaches for it.

prince-deriv: ok I understand the requirements. I think this is the only way to avoid reimplementation on each project.

lib/components/Snackbar/snackbar.stories.tsx
```diff
@@ -37,103 +59,167 @@ const meta = {
...
-        icon: "",
-        hasCloseButton: false,
-        actionText: "Action",
-        onActionClick: fn(),
-    },
+    return (
+        <div style={{ display: "flex", justifyContent: "center" }}>
+            <Button
+                size="lg"
+                label={"Add Snackbar"}
```
maryia-deriv, line 154:
```suggestion
                label="Add Snackbar"
```
nit-pick :)

maryia-deriv (changes requested):
the rest LGTM 👍

maryia-deriv (approved):

prince-deriv (approved):

jia-deriv (approved):

balakrishna-deriv (approved):

PR #76, opened 2024-04-22
Jia/QUILL-1524/Chips

jia-deriv:


lib/components/Chip/dropdown-chip-single-select/dropdown-chip-single-select.scss
```diff
@@ -0,0 +1,48 @@
...
+
+.leave {
+    transition: transform var(--core-motion-duration-200)
+        var(--core-motion-ease-400);
+    transform-origin: top;
+}
+
+.leave-from {
+    transform: scaleY(1);
+}
```
prince-deriv, line 22: Any chance that this can be reused on other components? If yes we can add it as part of our `custom` scss files.

prince-deriv (approved):

lib/components/Chip/dropdown-chip-single-select/dropdown-chip-single-select.scss
jia-deriv: This will be the item list in the Atom. I'm trying to find a way to move it to Atom folder with headless ui supported format.

PR #80, opened 2024-04-23
Prince/QUILL-1592/Update Quill Token

prince-deriv:
**Changes:**

- Updated `token variables` based on new `Design Tokens`
- Removed duplication of styles in each of component's CSS files. 
- Updated the approach to load `light` and `dark` theme styles. 
- Updated the old imports


lib/providers/theme/themeProvider.tsx
```diff
@@ -24,6 +24,21 @@ export const ThemeProvider = ({ children, theme }: ThemeProviderProps) => {
...
+            link.href = "/lib/styles/quill/light.scss";
+        } else {
+            link.href = "/lib/styles/quill/dark.scss";
+        }
+
+        document.head.appendChild(link);
+
+        return () => {
+            document.head.removeChild(link);
+        };
```
maryia-deriv, line 41: I noticed that this theme provider doesn't detect themes correctly, and with these changes, it's always dark for me.
Even though '**light theme**' is selected in storybook, it's still **dark** perhaps because my MacOS system prefers dark:
Screenshot 2024-04-24 at 5 27 02 PM

Also, I would recommend avoiding working with DOM directly as React suggests, and not append stylesheets to it but instead use an approach that we have in deriv-app:
importing the stylesheet into an entry file of the application and switching between themes based on the classNames **'theme--light' & 'theme--dark'**.
Deriv-app example:
Screenshot 2024-04-24 at 5 35 09 PM

We can still have separate files for light and dark styles if you need but they would be nested under those classnames:
```
:root {
    .theme--light {
```
@prince-deriv  Is it possible?

maryia-deriv (changes requested):

lib/providers/theme/themeProvider.tsx
prince-deriv: @maryia-deriv  thanks for the check and approach, suggestion. I have tried a closer approach and it did work

scripts/transformer.v1.js
```diff
@@ -0,0 +1,537 @@
...
+    getTokenGroup = (group) =>
+        Object.fromEntries(
+            Object.entries(this.designTokens).filter(([key]) =>
+                group.includes(key),
+            ),
+        );
+
+    isObject = (item) => typeof item === "object";
+
+    mapSASSValues = () => {
```
maryia-deriv, line 312:
```suggestion
    mapSassValues = () => {
```
nit-pick)

lib/providers/theme/themeProvider.tsx
maryia-deriv: @prince-deriv 🥲 unfortunately, the issue with dark theme always enforced still persists for me.
I guess my system prefers dark, so ThemeProvider takes it and it doesn't care which theme is selected in the app.
Screenshot 2024-04-25 at 1 52 41 PM

prince-deriv: @maryia-deriv  for storybook it will be a different fix. We will have a PR for that. Can you try running the app `npm run dev` and see if the theme works for you properly?

scripts/transformer.v1.js
prince-deriv: Sure 👍

vite.config.ts
```diff
@@ -38,6 +37,7 @@ export default defineConfig({
             "@hooks": resolve(__dirname, "lib/hooks"),
             "@providers": resolve(__dirname, "lib/providers"),
             "@utils": resolve(__dirname, "lib/utils"),
+            "@styles": resolve(__dirname, "lib/styles"),
```
Nuzhy-Deriv, line 40: shall we remove the `@quill-bundle` in line 33 as it's no longer exist?

prince-deriv: yes @Nuzhy-Deriv  that's true, good catch!

Nuzhy-Deriv (approved):

Nuzhy-Deriv (approved):

jia-deriv (approved):

maryia-deriv:
The styles of my Notification components are currently broken in your test link. Please check  🙏

kate-deriv (approved):

PR #86, opened 2024-04-26
Meenu update color tokens

meenakshi-deriv:


lib/components/Button/dropdown/dropdown.scss
```diff
@@ -12,27 +12,43 @@
...
+            min-height: 32px;
+           }
+           &-md {
+            min-height: 32px;
+           }
+           &-lg {
+            min-height: 48px;
+           }
+           &-xl {
+            min-height: 48px;
```
prince-deriv, line 37: can use core tokens yeah?

meenakshi-deriv: yeah we can use

prince-deriv (approved):

PR #99, opened 2024-04-30
Maryia/test: Add default timezone to jest config + Notifications improvements

maryia-deriv:
- Add default timezone to jest config 
- NotificationItem improvement for iOS to hide buttons upon clicking outside
- Add support for initial notification objects passed into useNotifications hook without ids: NotificationsProvider will add unique ids if they are missing.

### DIFF
--- a/globalSetup.js
+++ b/globalSetup.js
@@ -1,0 +1,4 @@
+export default async () => {
+    // eslint-disable-next-line no-undef
+    process.env.TZ = "UTC";
+};
--- a/lib/components/Notification/item/index.tsx
+++ b/lib/components/Notification/item/index.tsx
@@ -15,8 +15,10 @@ export interface NotificationItemProps
         | "onSwipeUp"
     > {
     className?: string;
-    type?: (typeof TYPE)[keyof typeof TYPE];
     isMobile?: boolean;
+    onShowingButtons?: () => void;
+    showButtons: boolean;
+    type?: (typeof TYPE)[keyof typeof TYPE];
 }
 
 const DIRECTION = {
@@ -32,6 +34,8 @@ const NotificationItem = ({
     onClick,
     onClose,
     onMarkAsRead,
+    onShowingButtons,
+    showButtons,
     ...rest
 }: NotificationItemProps) => {
     const [isDeleted, setIsDeleted] = useState(false);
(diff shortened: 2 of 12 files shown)


maryia-deriv:
not ready for merging yet

globalSetup.js
```diff
@@ -0,0 +1,4 @@
+export default async () => {
+    // eslint-disable-next-line no-undef
```
maryia-deriv, line 2: has to disable the 'process in not defined' error since it's defined for `jest` despite it's not defined for `vite`. The code's working properly.

lib/components/Notification/item/index.tsx
```diff
@@ -68,6 +68,7 @@ const NotificationItem = ({
                 "notification__item-wrapper",
                 isDeleted && "deleted",
             )}
+            onBlur={() => setShouldShowButtons(false)}
```
maryia-deriv, line 71: moved `onBlur` from anchor to `div` since `onBlur` doesn't work on anchors on iOS

lib/components/Notification/base.tsx
```diff
@@ -37,129 +37,135 @@ const swipeConfig = {
...
+                            onClick={handleClose}
+                        >
+                            <LabelPairedTrashSmBoldIcon />
+                        </button>
+                    </div>
+                )}
+            </a>
+        );
+    },
+);
```
maryia-deriv, line 171: in this file I only added `forwardRef` and `ref`

kate-deriv (approved):

lib/components/Notification/items-list/index.tsx
```diff
@@ -20,6 +20,7 @@ const NotificationItemsList = ({
     onClose,
     onMarkAsRead,
 }: NotificationItemsListProps) => {
+    const [activeItemId, setActiveItemId] = useState("");
```
maryia-deriv, line 23: using state for the list on Notifications in order to keep buttons open only on the latest active item where the buttons have been open last since on iOS, `onBlur` doesn't let us achieve it because it fires on the root `div` of the list instead of every notification. On other devices, `onBlur` fires on every notification.

prince-deriv (approved):

kate-deriv (approved):

PR #100, opened 2024-04-30
DTRA/ Kate / Update design tokens and refactor key events

kate-deriv:
- Update tokens
- Created _lib/utils/common-utils.ts_ file with strings for checking `e.key`
- Chore fixes (semicolons, moving `{...rest}` in order not to rewrite props and etc)

### DIFF
--- a/lib/utils/common-utils.ts
+++ b/lib/utils/common-utils.ts
@@ -1,0 +1,7 @@
+export const KEY = {
+    ARROW_LEFT: "ArrowLeft",
+    ARROW_RIGHT: "ArrowRight",
+    ENTER: "Enter",
+    SPACE: " ",
+    TAB: "Tab",
+};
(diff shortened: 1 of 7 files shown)


lib/utils/common-utils.ts
```diff
@@ -0,0 +1,7 @@
+export const KEY = {
+    ARROW_LEFT: "ArrowLeft",
+    ARROW_RIGHT: "ArrowRight",
+    ENTER: "Enter",
+    SPACE: " ",
+    TAB: "Tab",
+};
```
maryia-deriv, line 7: nit-pick: please add a line break in the end, so that git doesn't complain with  ⊖ sign :)

maryia-deriv (approved):

henry-deriv (approved):

prince-deriv (approved):

PR #103, opened 2024-05-02
Meenu/[QUILL-1609] Input dropdown component

meenakshi-deriv:
Input Dropdown component:

This component uses `combobox` from headless UI that handles autocomplete as well. To use normal dropdown without autocomplete we can pass the `readOnly` prop. It opens dropdown on clicking text field. `Keyboard` events also supported, like `onKeyDown` and mouse enter. Enable `floating label ` by passing label prop with required and field marker. Alignment and size of dropdown items can be controlled by passing `textAlign` and `inputSize` property. 

Screenshot 2024-05-10 at 2 33 53 PM
Screenshot 2024-05-10 at 2 34 13 PM
Screenshot 2024-05-10 at 2 48 07 PM


lib/components/Input/base/index.tsx
```diff
@@ -127,11 +129,8 @@ const Input = forwardRef<HTMLInputElement, InputProps>(
                             {rightIcon}
                         </span>
                     )}
-                    {triggerActionIcon && (
-                        <>
-                            {triggerActionIcon}
-                        </>
-                    )}
+                    {dropdownIcon && dropdownIcon}
```
henry-deriv, line 132: btw you can pass dropdown Icon into triggerActionIcon prop. I made the name generic so any icon can be passed in. No need to define a new prop 👍

meenakshi-deriv: Thanks @henry-deriv , i'll update it

prince-deriv (approved):

PR #108, opened 2024-05-03
DTRA-1162/ Kate / Fix: remove prevent default for focus

kate-deriv:
- Removed `e.preventDefault();` from `keyDown` handler as it prevent `focus` from changing target.

https://github.com/deriv-com/quill-ui/assets/121025168/b990150b-1246-4604-b209-15092f617dce

### DIFF
--- a/lib/components/SelectionControl/checkbox/checkbox-single/index.tsx
+++ b/lib/components/SelectionControl/checkbox/checkbox-single/index.tsx
@@ -89,7 +89,8 @@ export const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
 
         const onKeyDown: React.KeyboardEventHandler<HTMLInputElement> = (e) => {
             e.stopPropagation();
-            if (e.key === KEY.ENTER) onChangeHandler(e);
+            if (e.key === KEY.SPACE) e.preventDefault();
+            if (e.key === KEY.ENTER || e.key === KEY.SPACE) onChangeHandler(e);
         };
 
         const determinateIcon = isChecked ? (
--- a/lib/components/SelectionControl/checkbox/checkbox-single/index.tsx
+++ b/lib/components/SelectionControl/checkbox/checkbox-single/index.tsx
@@ -89,8 +89,6 @@ export const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
 
         const onKeyDown: React.KeyboardEventHandler<HTMLInputElement> = (e) => {
             e.stopPropagation();
-            e.preventDefault();
-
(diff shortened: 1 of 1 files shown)


prince-deriv (approved):

maryia-deriv (approved):

akmal-deriv (approved):

henry-deriv (approved):

lib/components/SelectionControl/checkbox/checkbox-single/index.tsx
```diff
@@ -89,8 +89,6 @@ export const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
 
         const onKeyDown: React.KeyboardEventHandler<HTMLInputElement> = (e) => {
             e.stopPropagation();
-            e.preventDefault();
```
prince-deriv, line 92: Check the original reason why we needed to add this in the first place.

kate-deriv: There is a well-known behavior, that `spacebar` (from keyboard) is triggering `click` event on such elements like input ([Stackoverflow](https://stackoverflow.com/questions/27878940/spacebar-triggering-click-event-on-checkbox) topic, [article](https://www.w3.org/TR/WCAG20-TECHS/SCR35)). That was made for accessibility purposes: browsers use the click event even if its triggered from the keyboard in the event where a mouse isn't present. 
Because of that, when user hit `Space`, `onChange` was triggered twice (as a `keyboard` and as a `mouse` event). I didn't want this behavior and add `e.preventDefault();` inside of `onKeyDown` in order not to trigger `onChange` twice. That was the original reason.
But it turned out, that `e.preventDefault();` also disabled `focus` from jumping to another element when user is pressing 'Tab', so after all I decided to remove `e.preventDefault();` and to remove `e.key === KEY.SPACE` from check, because of the reasons mentioned above (hitting space will trigger `onChange` anyway)

prince-deriv: Can we detect if the even came from the spacebar and that's the only time we will fire `e.preventDefault();`

kate-deriv: Yep, good catch, thanks! It should work, let me add

kate-deriv: Sorry for my stupidness)

prince-deriv: You did a great job checking the UI events, thanks for that!

prince-deriv (approved):

PR #110, opened 2024-05-03
Meenu/[QUILL-786] Modal overlay component

meenakshi-deriv:
Modal overlay (desktop only)

This component is used only for desktop and it accepts props for button, image, icon or any content. Implementation is similar to modal button except the position of the container. The main container is fixed to the center of the page. 

Screenshot 2024-05-03 at 4 05 14 PM
Screenshot 2024-05-03 at 4 05 33 PM

### DIFF
--- a/lib/components/Modal/index.ts
+++ b/lib/components/Modal/index.ts
@@ -1,1 +1,2 @@
-export * from './bottom';
+export * from "./bottom";
+export * from "./overlay";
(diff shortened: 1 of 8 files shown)


prince-deriv:
@meenakshi-deriv  @kate-deriv  let's find a way to have a `base` component for `modal` 
so we can avoid duplication of code on different variants.

meenakshi-deriv:
Fixed in another PR

PR #112, opened 2024-05-05
Nuzhy/[QUILL-1637]/Input Field Update Tokens

Nuzhy-Deriv:
- update newly assigned tokens for input text field base component
- implemented character-count with maxLength prop

### DIFF
--- a/lib/components/Input/base/index.tsx
+++ b/lib/components/Input/base/index.tsx
@@ -29,7 +29,7 @@ export interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
     disabled?: boolean;
     variant?: Variants;
     message?: string;
-    characterCounter?: boolean;
+    showCharacterCounter?: boolean;
     maxLength?: number;
     textAlignment?: TextAlignments;
     label?: ReactNode;
@@ -62,7 +62,7 @@ const Input = forwardRef<HTMLInputElement, InputProps>(
             placeholder = "",
             leftIcon,
             message,
-            characterCounter,
+            showCharacterCounter,
             maxLength,
             textAlignment = "left",
             label,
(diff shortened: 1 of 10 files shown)


jia-deriv (approved):

meenakshi-deriv (approved):

lib/components/Input/base/index.tsx
```diff
@@ -131,36 +158,35 @@ const Input = forwardRef<HTMLInputElement, InputProps>(
...
+                        <span className="message__container__text">
+                            {message}
+                        </span>
+                        {characterCounter && maxLength && (
+                            <span className="message__container__text">
+                                {inputValue.length}/{maxLength}
+                            </span>
+                        )}
+                    </div>
+                )}
```
prince-deriv, line 189: **Quick question:** So even if let's say we don't have `message` we still render the UI? Is it because we need it for the UI consistency?

lib/components/Input/base/index.tsx
```diff
@@ -18,8 +28,9 @@ export interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
     status?: Status;
     disabled?: boolean;
     variant?: Variants;
-    leftStatusMessage?: string;
-    rightStatusMessage?: string;
+    message?: string;
+    characterCounter?: boolean;
```
prince-deriv, line 31: Maybe we can make the props a little bit clearer. 

**showCharacterCount:** Indicates whether to display the character count on the input.
**enableCharacterCount:** Enables the character count feature for the input.
**displayCharacterCount:** Controls the visibility of the character count on the input.

We can choose from this based on what is the real purpose of the prop.

lib/components/Input/base/index.tsx
```diff
@@ -33,6 +44,12 @@ const statusIconColors = {
     error: "status-icon--error",
 };
 
+const statusIcon = {
+    success: <StandaloneCircleCheckBoldIcon iconSize="sm" />,
+    error: <StandaloneTriangleExclamationBoldIcon iconSize="sm" />,
+    neutral: null,
```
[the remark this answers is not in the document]
Nuzhy-Deriv: yes @prince-deriv. Earlier implementation to accept icon from props. As checked  on the figma, success, error icons are always same.

prince-deriv (approved):

PR #120, opened 2024-05-08
Akmal / feat: add input with steppers

akmal-deriv:


lib/components/Input/text-field-with-steppers/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+import {
+    LabelPairedPlusSmRegularIcon,
+    LabelPairedMinusSmRegularIcon,
+} from "@deriv/quill-icons";
+import "./text-field-with-steppers.scss";
+
+export interface TextFieldWithSteppersProps extends ComponentProps<typeof Input> {
+    unit?: string;
+    decimals?: number;
+    unitPlacement?: 'left' | 'right';
```
prince-deriv, line 14: can we use the generic type `TLeftOrRight`

lib/components/Input/text-field-with-steppers/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+                <Input
+                    {...props}
+                    ref={ref}
+                    triggerActionIcon={steppersSectionRight}
+                    leftIcon={textAlignment === 'center' && steppersSectionLeft}
+                    type="number"
+                    value={`${getFormatValue(value as number)}`}
+                    placeholder={placeholder.toLocaleString("en", { minimumFractionDigits: decimals })}
+                    leftPlaceholder={unitPlacement === "left" ? unit : undefined}
+                    rightPlaceholder={unitPlacement === "right" ? unit : undefined}
```
prince-deriv, line 82:
```suggestion
                    leftPlaceholder={unitPlacement === "left" && unit}
                    rightPlaceholder={unitPlacement === "right" && unit}
```

akmal-deriv: This doesn't work because in this case the type becomes false and I don't want to add boolean as a type for left and right placeholders since it's kinda meaningless

prince-deriv: okay I see. the type is `string`

lib/components/Input/text-field-with-steppers/index.tsx
akmal-deriv: Yes, I didn't think there's already such type

prince-deriv (approved):

lib/components/Input/base/index.tsx
```diff
@@ -186,6 +209,21 @@ const Input = forwardRef<HTMLInputElement, InputProps>(
...
+                                        as="span"
+                                        className={clsx(
+                                            "quill-input-label__label",
+                                            "quill-input-label__label--right",
+                                            `quill-input-label__label--disabled`,
+                                        )}
+                                    >
+                                        {rightPlaceholder}
+                                    </Text>
+                                )}
```
maryia-deriv, line 226: it looks like leftPlaceholder and rightPlaceholder have exactly the same logic, can we move it out and reuse for both? It can be a small Placeholder component, which would accept the content for any placeholder? :)

maryia-deriv (changes requested):

lib/components/Input/text-field-with-steppers/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+        const { value, onChange, disabled, textAlignment, placeholder = 0, unit = '', decimals = 2, unitPlacement = 'left' } = props;
+
+        const getFormatValue = (value: number) => {
+            let formatValue = value;
+
+            if (typeof formatValue === 'number') {
+                formatValue = parseFloat(formatValue.toFixed(decimals));
+            }
+
+            return formatValue;
```
kate-deriv, line 29: `return typeof formatValue === 'number' ? parseFloat(formatValue.toFixed(decimals)) : formatValue`
May be it will be shorter?

kate-deriv: Also, you determine the type of `value` as `number` ( `const getFormatValue = (value: number) =>`), do we need to check it again?

lib/components/Input/text-field-with-steppers/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+                        }
+                    }}>
+                    <LabelPairedPlusSmRegularIcon />
+                </button>
+            </>
+        );
+
+        const steppersSectionLeft = (
+            <>
+                <button className={clsx("quill-input-steppers-decrement", disabled && 'quill-input-steppers--disabled')}
```
kate-deriv, line 59: This code looks like first part of `steppersSectionRight`. is it possible to extract them and reuse?

akmal-deriv: There's no point or benefit of reusing this.

lib/components/Input/base/index.tsx
akmal-deriv: There's no point or benefit of creating any new component for this.

PR #123, opened 2024-05-09
Meenu/Updated close icon as button

meenakshi-deriv:


lib/components/Modal/base/index.tsx
```diff
@@ -145,10 +145,12 @@ export const Modal = ({
...
                     )
                 ) : (
-                    <LabelPairedXmarkMdBoldIcon
+                    <button
                         className="quill-modal__close-icon"
                         onClick={toggleHandler}
-                    />
+                    >
+                        <LabelPairedXmarkMdBoldIcon fill="var(--component-textIcon-normal-prominent)" />
+                    </button>
```
jia-deriv, line 153: If we use our quill button component will have the padding I guess.

meenakshi-deriv: we don't have this variant in quill button and it all comes with a border radius or underline. so i used custom button

lib/components/Modal/modal.scss
```diff
@@ -82,7 +82,11 @@
     &__close-icon {
         position: absolute;
         cursor: pointer;
-        top: var(--core-spacing-1600);
+        top: var(--core-spacing-1400);
+        width: var(--core-spacing-2400);
+        padding: var(--core-spacing-600) var(--core-spacing-800);
+        justify-content: center;
+        display: flex;
         right: var(--core-spacing-1600);
```
jia-deriv, line 90:
```suggestion
        right: var(--core-spacing-800);
```

lib/components/Modal/modal.scss
```diff
@@ -82,7 +82,11 @@
     &__close-icon {
         position: absolute;
         cursor: pointer;
-        top: var(--core-spacing-1600);
+        top: var(--core-spacing-1400);
```
jia-deriv, line 85:
```suggestion
        top: var(--core-spacing-800);
```

image

jia-deriv (approved):

prince-deriv (approved):

PR #126, opened 2024-05-10
Meenu/fix: added dropdown to base input

meenakshi-deriv:


lib/components/Input/base/index.tsx
```diff
@@ -164,6 +173,7 @@ const Input = forwardRef<HTMLInputElement, InputProps>(
 
                     {triggerActionIcon && <>{triggerActionIcon}</>}
                 </div>
+                {hasDropdown && isDropdownOpen && !disabled && children}
```
kate-deriv, line 176: I think children from dropdown should have some kind of container with positioning (in order not to shift neighboring elements)

lib/components/Input/base/index.tsx
```diff
@@ -78,6 +82,10 @@ const Input = forwardRef<HTMLInputElement, InputProps>(
         useEffect(() => {
             setInputValue(value || "");
         }, [value]);
+        const [isDropdownOpen, setIsDropdownOpen] = useState(false);
+        const handleDropdown = () => {
```
kate-deriv, line 86: Are we going to add any animation for dropdown?

lib/components/Input/base/index.tsx
kate-deriv: + According to design, dropdown should have title (not only placeholder) and arrow icon, which will indicate if dropdown is expanded or collapsed. 
Taking in account all this maybe it's better to create a separate component, based on ./base/index.tsx?
Screenshot 2024-05-10 at 11 25 18 AM

kate-deriv, line 176: Btw dropdown is not hiding after appearing in testlink

PR #160, opened 2024-05-22
DTRA / Kate / Fix hover and active state

kate-deriv:
Changed ccs for styling state for the button.

### DIFF
--- a/lib/components/Atom/Calendar/date-picker.scss
+++ b/lib/components/Atom/Calendar/date-picker.scss
@@ -26,6 +26,7 @@
             }
 
             button:enabled:hover:is(.react-calendar__navigation__arrow),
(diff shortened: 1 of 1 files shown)


lib/components/Atom/Calendar/date-picker.scss
```diff
@@ -20,12 +20,44 @@
             margin-bottom: var(--semantic-spacing-general-md);
 
             button:enabled:hover,
-            button:enabled:focus {
+            button:enabled:focus,
+            button:disabled {
+                background-color: unset;
+            }
+
+            button:enabled:focus:is(.react-calendar__navigation__arrow),
```
kate-deriv, line 28: Unfortunately, in the library this style were written with tag selectors, so in order to overwrite them I was forced to do the same

prince-deriv (approved):

PR #197, opened 2024-06-05
Akmal / feat: improve input with steppers

akmal-deriv:
- improve input with steppers


lib/components/Input/base/base.scss
```diff
@@ -118,7 +118,7 @@
             margin-left: var(--core-spacing-200);
         }
         &--disabled {
-            opacity: var(--core-opacity-600);
+            color: var(--component-field-label-color-disabled) !important;
```
prince-deriv, line 121: What is the style we are trying to override and we needed to add `important`. Maybe we can handle the class structure to handle this?

akmal-deriv: There's default color assigned to text component that I'm trying to overwrite. It always has a priority over any color I pass. I'm thinking it should be fixed by adding color prop to text component or not enforcing any default color that can't be changed without higher priority.

lib/components/Input/text-field-with-steppers/index.tsx
```diff
@@ -7,55 +7,60 @@ import {
...
+        const { value, onChange, disabled, textAlignment, placeholder = 0, unitLeft = '', unitRight = '', decimals = 2, ...rest } = props;
 
         const getFormatValue = (value: number) => parseFloat(value.toFixed(decimals));
 
         const steppersSectionRight = (
             <>
-                {textAlignment !== 'center' && (<button className={clsx("quill-input-steppers-decrement", disabled && 'quill-input-steppers--disabled')}
+                {textAlignment !== 'center' && (<button className={clsx("quill-input-steppers-button", "quill-input-steppers-button--decrement")}
                     onClick={() => {
                         if (!disabled) {
```
kate-deriv, line 27: I'm not sure, bit if button is `disabled` user can't click, so this check might be additional

lib/components/Input/base/base.scss
akmal-deriv: Ok, I noticed there's already color property in the text component, let me try to use that to overwrite it

lib/components/Input/text-field-with-steppers/index.tsx
```diff
@@ -7,55 +7,60 @@ import {
...
+                    >
                     <LabelPairedPlusSmRegularIcon />
                 </button>
             </>
         );
 
         const steppersSectionLeft = (
             <>
-                <button className={clsx("quill-input-steppers-decrement", disabled && 'quill-input-steppers--disabled')}
+                <button className={clsx("quill-input-steppers-button", "quill-input-steppers-button--decrement")}
```
kate-deriv, line 54: Maybe if `jsx` with `button`  is used in multiple places we can create small reusable component?
Like component, which will take as props `className`, `icon` and `- 1` or `+ 1` for `onClick` function calculations?

lib/components/Input/text-field-with-steppers/index.tsx
kate-deriv: Please, ignore if irrelevant

lib/components/Input/text-field-with-steppers/text-field-with-steppers.scss
```diff
@@ -1,19 +1,19 @@
...
+            background-color: var(--core-color-opacity-black-100);
+        }
+        &:active:not(:disabled) {
+            background-color: var(--core-color-opacity-black-200);
+        }
+        &:disabled {
+            cursor: not-allowed;
+        }
     }
 }
```
maryia-deriv, line 19: just a nit-pick:) Do you mind adding a line break at the end of the file to stop Git from complaining about it with ⊖ sign?

lib/components/Input/base/base.scss
akmal-deriv: I removed this style

lib/components/Input/text-field-with-steppers/index.tsx
akmal-deriv: It's totally not worth it, it's used 2 times in its existence

lib/components/Input/text-field-with-steppers/text-field-with-steppers.scss
akmal-deriv: Sure, added

prince-deriv (approved):

kate-deriv (approved):

ahmadtaimoor-deriv (approved):

PR #206, opened 2024-06-06
Jia/_/icon-button

jia-deriv:
Added `IconButton` component
image


lib/components/Button/base/index.tsx
```diff
@@ -83,7 +85,7 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
                         className="quill-button__loader-icon"
                     />
                 )}
-                {children && <div>{children}</div>}
```
jia-deriv, line 86: @meenakshi-deriv is there any reason we have a div to wrap the children? Because of this, the icon are not aligned correctly. After I remove the div Button component is still working correctly.

Nuzhy-Deriv (approved):

lib/components/Button/base/index.tsx
```diff
@@ -10,7 +10,7 @@ import {
     LabelPairedLoaderLgRegularIcon,
     LabelPairedLoaderMdRegularIcon,
     LabelPairedLoaderSmRegularIcon,
-    IconTypes
+    IconTypes,
```
prince-deriv, line 13: Why do we add comma here?

jia-deriv: This comma is added by Prettier

lib/components/Button/icon-button/index.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { forwardRef } from "react";
+import Button from "../base";
+import { IconButtonProps } from "../types";
+
+export const IconButton = forwardRef<HTMLButtonElement, IconButtonProps>(
+    ({ ...rest }, ref) => {
+        return <Button iconButton={true} ref={ref} {...rest} />;
```
prince-deriv, line 7:
```suggestion
        return <Button iconButton ref={ref} {...rest} />;
```

lib/components/Button/base/index.tsx
```diff
@@ -28,7 +28,7 @@ const dropdownIcons = {
     lg: LabelPairedChevronDownLgBoldIcon,
     xl: LabelPairedChevronDownXlBoldIcon,
 };
-export const loaderIcons: Record<'sm' | 'md' | 'lg' | 'xl', IconTypes> = {
+export const loaderIcons: Record<"sm" | "md" | "lg" | "xl", IconTypes> = {
```
prince-deriv, line 31: use the size types in `@types`

prince-deriv (approved):

PR #216, opened 2024-06-13
Jia/_/storybook-enhancement

jia-deriv:


lib/components/Link/link.stories.tsx
```diff
@@ -14,7 +14,7 @@ const meta = {
     title: "Components/Links/Link",
     component: Link,
     parameters: {
-        layout: "centered",
+        // layout: "centered",
```
prince-deriv, line 17: Can we remove instead?

lib/components/Tag/custom-tag.stories.tsx
```diff
@@ -52,9 +53,10 @@ const meta = {
             control: "boolean",
         },
         color: {
-            table: {
-                disable: true,
-            },
+            // table: {
+            //     disable: true,
+            // },
```
prince-deriv, line 58: let's remove this if not needed

prince-deriv (changes requested):

lib/components/Link/link.stories.tsx
jia-deriv: Sure

lib/components/Tag/custom-tag.stories.tsx
jia-deriv: Sure

prince-deriv (approved):

PR #218, opened 2024-06-14
Jia/QUILL-1758/Create a standalone snackbar component

jia-deriv:
image
Create a standalone snackbar without relying on hooks


meenakshi-deriv (approved):

lib/components/Snackbar/snackbar.tsx
```diff
@@ -34,16 +36,17 @@ export const Snackbar = ({
...
             clearTimeout(timerRef.current ?? "");
         };
     }, [message]);
 
-    const handleClose = (delay: number = 100) => {
+    const handleClose = () => {
+        onCloseAction?.();
         if (timerRef) clearTimeout(timerRef.current ?? "");
-        snackbarDelayedRemove(id, delay);
+        id && snackbarDelayedRemove(id, 100);
```
prince-deriv, line 49: why aren't we using the `delay` anymore? We are allowing the consumer to add their own delay while we are setting a default delay of `100`

jia-deriv: Because I noticed delay is not passing into the function then I remove it. I'll add this as a props for consumer

prince-deriv (approved):

PR #237, opened 2024-06-25
Jia/_/add tab component initial active tab props

jia-deriv:
- added `selectedTabIndex` props to control the current active tab.
- added `onChangeTab` to have a callback on the selectedTabIndex value changed

### DIFF
--- a/lib/components/Tab/mocks/index.tsx
+++ b/lib/components/Tab/mocks/index.tsx
@@ -6,7 +6,7 @@ const MockTab = ({
     size,
     iconPosition,
     contentStyle,
-    initialActiveTab,
+    selectedTabIndex,
     ...props
 }: TabProps) => {
     return (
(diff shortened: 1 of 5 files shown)


prince-deriv (approved):

lib/components/Tab/mocks/index.tsx
```diff
@@ -2,13 +2,23 @@ import Tab from "@components/Tab";
...
+}: TabProps) => {
     return (
         <Tab.Container
             id="test"
             size={size}
             iconPosition={iconPosition}
             contentStyle={contentStyle}
+            selectedTabIndex={selectedTabIndex}
+            onChangeTab={(value) => {
+                console.log(value);
```
kate-deriv, line 20: Do we need this console.log?

prince-deriv: I think this is for storybook

PR #290, opened 2024-07-10
nuzhy/_/adding tooltip to checkbox info icon

Nuzhy-Deriv:
- exported tooltip component
- added Tooltip to the checkbox info icon to show message on hover
- replaced `showInfoIcon ` prop with `infoIconMessage `
- set input type `number` to `text`


lib/components/SelectionControl/checkbox/checkbox-single/index.tsx
```diff
@@ -164,13 +165,18 @@ export const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
...
-                        className={clsx(
-                            "quill-checkbox__info-icon",
-                            infoIconClassName,
-                        )}
-                    />
+                {infoIconMessage && (
+                    <Tooltip
+                        shouldCloseToolTipOnMouseLeave
+                        tooltipContent={infoIconMessage}
+                    >
```
jia-deriv, line 172: Do we need to allow consumer to change the tooltip position or is it fixed in the design?

Nuzhy-Deriv: it fixed as per figma in smart trader. if needed in future, tooltip position can be added.

lib/components/Tooltip/index.tsx
```diff
@@ -146,4 +144,4 @@ export const Tooltip = <T extends AsElement>({
     );
 };
 
-Tooltip.displayName = "Tooltip";
```
prince-deriv, line 149: You can actually keep this and help us on debugging.

jia-deriv (approved):

prince-deriv (approved):

PR #302, opened 2024-07-16
Nuzhy/_/update input id

Nuzhy-Deriv:
- updated the input base id to uniqueID based on counter if the Id is not present.


jia-deriv (approved):

henry-deriv (approved):

lib/components/Input/textarea/index.tsx
```diff
@@ -66,6 +67,8 @@ export const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(
                 ? statusIcon[status]
                 : rightIcon;
 
+        const id = uuid();
```
balakrishna-deriv, line 70: can we use generateUniqueId here as well?

package.json
```diff
@@ -123,7 +123,8 @@
     "jest": {
         "testEnvironment": "jest-environment-jsdom",
         "setupFilesAfterEnv": [
-            "<rootDir>/test/setup.test.js"
+            "<rootDir>/test/setup.test.js",
+            "<rootDir>/test/mock-uuid.js"
```
balakrishna-deriv, line 127: can we remove the mock-uuid?

lib/components/Input/textarea/index.tsx
Nuzhy-Deriv: oh yeah. missed this. updating it

balakrishna-deriv (approved):

PR #303, opened 2024-07-16
DTRA-1478 / Kate / Action sheet component improvement

kate-deriv:
Refactored `ActionSheet` component from Quill UI:
1) Unmount `ActionSheet` component (not just hide with css) when props `is_open` is falsy. In DTrader for such purpose we are using `CSSTransition`, but as it required new npm package installation I decided to do it with `setTimeout`.
2) Increase draggable area of handle bar in order to improve UX (by adding pseudo-element with height 48px for handle bar). Swipable area is marked with red color on the screenshot).
3) Add npm package `@types/lodash.debounce`
4) In _lib/components/ActionSheet/portal/index.tsx_ added props passing (`toggleHandler`) from parent component to Children (only for Footer). I was forced to do it, because Footer buttons are among children.

Screenshot 2024-07-16 at 5 10 00 PM

https://github.com/user-attachments/assets/d9bead97-863b-41af-a295-6f5cc7354b0d

Screenshot 2024-07-16 at 5 52 05 PM

### DIFF
--- a/lib/components/ActionSheet/portal/__tests__/portal.test.tsx
+++ b/lib/components/ActionSheet/portal/__tests__/portal.test.tsx
@@ -44,8 +44,7 @@ jest.mock("usehooks-ts", () => ({
     })),
 }));
 
-const mock_loadash = jest.fn((fn) => fn);
-jest.mock("lodash/debounce", () => mock_loadash);
+jest.mock("lodash.debounce", () => jest.fn((fn) => fn));
 
 describe("<ActionSheet.Portal/>", () => {
     it('should set the data-state attribute to "open" when the show is true', async () => {
--- a/lib/components/ActionSheet/portal/__tests__/portal.test.tsx
+++ b/lib/components/ActionSheet/portal/__tests__/portal.test.tsx
@@ -9,14 +9,14 @@ import userEvent from "@testing-library/user-event";
 import { RootPosition, RootProps } from "../../types";
 import ActionSheet from "@components/ActionSheet";
 
-const dialog_state = {
-    open: "open",
-    close: "close",
+const DIALOG_STATE = {
+    OPEN: "open",
+    CLOSE: "close",
 };
 
-const button_name = {
-    trigger: "Trigger",
-    close: "Close",
+const BUTTON_NAME = {
+    TRIGGER: "Trigger",
+    CLOSE: "Close",
 };
 
 const overlay = "dt-actionsheet-overlay";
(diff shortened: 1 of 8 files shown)


lib/components/ActionSheet/portal/__tests__/portal.test.tsx
```diff
@@ -31,91 +44,143 @@ jest.mock("usehooks-ts", () => ({
     })),
 }));
 
+const mock_loadash = jest.fn((fn) => fn);
+jest.mock("lodash/debounce", () => mock_loadash);
```
maryia-deriv, line 48: nit-pick)
```suggestion
jest.mock('lodash.debounce', () => jest.fn(fn => fn));
```

lib/components/ActionSheet/portal/index.tsx
```diff
@@ -13,6 +22,14 @@ interface PortalProps extends ComponentProps<"div"> {
     fullHeightOnOpen?: boolean;
 }
 
+type TCustomConstructor = React.JSXElementConstructor<{
+    handleClose: () => void;
+}>;
+type TChild = React.ReactElement<
+    { handleClose: () => void },
```
maryia-deriv, line 29: nit-pick: just in order to not write the object twice :)
```
type TFooterChildProps = { handleClose: () => void };
type TFooterJSXConstructor = React.JSXElementConstructor<TFooterChildProps>;
type TFooterChild = React.ReactElement<TFooterChildProps, TFooterJSXConstructor>;
```

maryia-deriv (changes requested):
just a couple of nit-picks, the rest LGTM 👍

ahmadtaimoor-deriv (approved):

jia-deriv (approved):

lib/components/ActionSheet/portal/__tests__/portal.test.tsx
kate-deriv: Oh, yep, sorry) Will apply

lib/components/ActionSheet/portal/index.tsx
```diff
@@ -27,38 +43,67 @@ const Portal = forwardRef<HTMLDivElement, PortalProps>(
...
+
+        const [isVisible, setIsVisible] = useState(show);
+        const animationTimerRef = useRef<ReturnType<typeof setTimeout>>();
+
+        const toggleHandler = debounce(() => {
+            setIsVisible(!isVisible);
+            animationTimerRef.current = setTimeout(() => handleClose?.(), 500);
+        }, 200);
+
+        const childrenWithProps = React.Children.map(children, (child) => {
```
kate-deriv, line 55: I'm mapping through children. If it's a `Footer` (`ActionSheet.Footer`), then I'm cloning it and passing `handleClose` function from Parent (Portal ). Now, `Footer` won't use function from context, but get a new one  from prop. Other props won't be overwritten. 
Old function, which `Footer` was getting from context: `handleClose`.
New one, which `Footer` gets from props from the Parent:        
```
 const toggleHandler = debounce(() => {
            setIsVisible(!isVisible);
            animationTimerRef.current = setTimeout(() => handleClose?.(), 500);
        }, 200);
```

lib/components/ActionSheet/portal/index.tsx
kate-deriv: Thank you

henry-deriv (approved):

prince-deriv (approved):

maryia-deriv (approved):

PR #318, opened 2024-07-25
Jia/_/fix input textfield 0 value

jia-deriv:


lib/components/Chip/base/chip.scss
```diff
@@ -6,6 +6,7 @@
     border-width: var(--component-chip-border-width);
     border-color: var(--component-chip-border-color);
     border-radius: var(--component-chip-border-radius);
+    background-color: transparent;
```
jia-deriv, line 9: To fix the default button background color appearing on chip component.

prince-deriv (approved):

PR #325, opened 2024-08-01
Ahmad/Adding allowSigns flag in the textfield 

ahmadtaimoor-deriv:
This pull request introduces the allowSigns flag to the TextField component, enhancing its functionality to support input with sign characters, specifically positive (+) and negative (-) signs. This addition caters to use cases where numeric inputs may need to disable positive and negative values

### DIFF
--- a/lib/components/Input/text-field/text-field.stories.tsx
+++ b/lib/components/Input/text-field/text-field.stories.tsx
@@ -46,6 +46,7 @@ const meta = {
         required: false,
         show_counter: false,
         allowDecimals: false,
+        allowSign: true,
     },
     argTypes: {
         type: {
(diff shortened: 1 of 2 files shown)


jia-deriv (approved):

lib/components/Input/text-field/text-field.stories.tsx
```diff
@@ -58,6 +59,10 @@ const meta = {
         allowDecimals: {
             description: "input type number to allow decimal values",
         },
+        allowSign: {
+            description:
+                "input type number to allow negative and poisitive sign only on decimal value",
```
prince-deriv, line 64:
```suggestion
                "input type number to allow negative and positive sign only on decimal value",
```

PR #355, opened 2024-08-06
Meenu/Breadcrumb component

meenakshi-deriv:


lib/components/Breadcrumbs/base/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+            const remainingLinks: LinkProps[] = links.filter(
+                (item) => !new Set(limitedLinks).has(item),
+            );
+
+            setDropdownLinks(remainingLinks);
+        }
+    }, [links]);
+
+    return (
+        <div className={clsx("breadcrumb--container", className)}>
```
prince-deriv, line 40: Can we make sure all components container has a classname starts with `quill-` so we can handle them accordingly in the future.

prince-deriv (approved):

prince-deriv:
Breadcrumb is not exported

PR #389, opened 2024-08-23
Jia/_/reusability of actionsheet modal component

jia-deriv:
- update modal component to use the same bottom sheet as action sheet for mobile screen. All the props and behaviour and ui should be same as previous.


lib/components/Modal/base/index.tsx
```diff
@@ -32,172 +33,130 @@ export interface ModalProps extends HTMLAttributes<HTMLDivElement> {
...
+            <ActionSheet.Portal
+                {...rest}
+                showHandlebar={showHandleBar}
+                handleBarPosition="absolute"
+                disableCloseOnOverlay={disableCloseOnOverlay}
+                portalId={portalId}
+            >
+                <ModalContent />
+            </ActionSheet.Portal>
+        </ActionSheet.Root>
```
jia-deriv, line 159: We can't make Modal component to fully action sheet because there is some different design for modal component like padding, h4 title (h5 in actionsheet), image header and etc. But the portal and handle swipe close can be use actionsheet portal, in this case we can remove duplicated swipe function in modal component.

Nuzhy-Deriv (approved):

balakrishna-deriv (approved):

PR #393, opened 2024-08-26
Meenu/fix: added lg size

meenakshi-deriv:
-added lg size for text field

Screenshot 2024-08-27 at 10 14 47 AM

Screenshot 2024-08-27 at 10 14 09 AM

### DIFF
--- a/lib/components/Input/base/base.scss
+++ b/lib/components/Input/base/base.scss
@@ -298,6 +298,12 @@
                 border-radius: var(--component-field-border-radius-md);
                 gap: var(--core-spacing-800);
             }
+            &--lg {
+                padding-inline: var(--component-field-spacing-padding-md);
+                height: 40px;
+                border-radius: var(--component-field-border-radius-md);
+                gap: var(--component-field-spacing-gap-sm); 
+            }
         }
 
         .input {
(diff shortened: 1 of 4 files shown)


lib/components/Input/base/base.scss
```diff
@@ -298,6 +298,12 @@
                 border-radius: var(--component-field-border-radius-md);
                 gap: var(--core-spacing-800);
             }
+            &--lg {
+                padding-inline: var(--component-field-spacing-padding-md);
+                height: 40px;
```
jia-deriv, line 303: Do we have a token for this?

meenakshi-deriv: Yeah we can, updated token.

lib/components/Input/base/base.scss
```diff
@@ -298,6 +298,12 @@
                 border-radius: var(--component-field-border-radius-md);
                 gap: var(--core-spacing-800);
             }
+            &--lg {
+                padding-inline: var(--component-field-spacing-padding-md);
+                height: var(--core-spacing-2000);
+                border-radius: var(--component-field-border-radius-md);
+                gap: var(--component-field-spacing-gap-sm); 
+            }
```
Nuzhy-Deriv, line 306: as checked the figma, newly added size is `md`. previous `md` has to changed to `lg` size. please check

PR #397, opened 2024-08-27
[DTRA] Maryia/feat: add isOpaque prop to Button component

maryia-deriv:
#### Added an optional `isOpaque` prop to Button component which will add a white background to the button when we need it to be non-transparent (opaque) in disabled state.

This is necessary because in deriv-app consumer app, devs are currently forced to add the white background to every single `Button` because according to design in DTrader most of the buttons are floating and should be opaque when disabled.

Issue:
Screenshot 2024-08-27 at 10 57 10 AM

Expected behavior:
Screenshot 2024-08-27 at 10 57 48 AM

### DIFF
--- a/lib/components/Button/base/index.tsx
+++ b/lib/components/Button/base/index.tsx
@@ -1,4 +1,4 @@
-import { forwardRef } from "react";
+import React, { forwardRef } from "react";
 import clsx from "clsx";
 import { ButtonProps } from "../types";
 import {
@@ -52,6 +52,7 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
             label,
             disabled,
             iconPosition = "start",
+            isOpaque,
             variant = "primary",
             iconButton = false,
             ...rest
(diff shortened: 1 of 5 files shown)


lib/components/Button/base/index.tsx
```diff
@@ -63,55 +64,70 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
...
+                            data-state={isDropdownOpen ? "open" : "close"}
+                            className={clsx(
+                                "quill-button__transform",
+                                isDropdownOpen &&
+                                    "quill-button__transform-rotate",
+                            )}
+                        />
+                    )}
+                </button>
+            </Wrapper>
```
maryia-deriv, line 130: i only wrapped the component with the Wrapper. Please check in github.dev using 'shift' + '.'

Nuzhy-Deriv (approved):

jia-deriv (approved):

meenakshi-deriv (approved):

kate-deriv (approved):

balakrishna-deriv (approved):

PR #399, opened 2024-08-28
Nuzhy/_/showChevron prop in button dropdown

Nuzhy-Deriv:
- added prop to button dropdown to show hide chevron


jia-deriv (approved):

meenakshi-deriv (approved):

lib/components/Button/base/index.tsx
```diff
@@ -45,6 +45,7 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
             children,
             isDropdownOpen,
             dropdown = false,
+            showChevron = true,
```
prince-deriv, line 48: since `showChevron` is `true` by default, can we make if `hideChevron` instead?

prince-deriv (approved):

PR #468, opened 2024-10-14
DTRA-2035 / Kate / Expand pagination functionality

kate-deriv:
- Added optional prop initialPage (stands for setting initial page)
- Made onClickPagination callback optional and removed it from array with dependencies as it cause infinite calls

### DIFF
--- a/lib/components/Pagination/index.tsx
+++ b/lib/components/Pagination/index.tsx
@@ -36,7 +36,7 @@ export const Pagination = ({
 
     useEffect(() => {
         onClickPagination?.({ currentPage, totalPageCount });
-    }, [currentPage, totalPageCount]);
+    }, [currentPage]);
 
     const goToNextPage = () => setCurrentPage((page) => page + 1);
 
--- a/lib/components/Pagination/index.tsx
+++ b/lib/components/Pagination/index.tsx
@@ -13,11 +13,12 @@ import "./pagination.scss";
 export const Pagination = ({
     contentPerPage,
     contentLength,
+    initialPage,
     onClickPagination,
     variant = "number",
     hideChevron,
 }: PaginationProps) => {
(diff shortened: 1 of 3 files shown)


lib/components/Pagination/index.tsx
```diff
@@ -13,11 +13,12 @@ import "./pagination.scss";
 export const Pagination = ({
     contentPerPage,
     contentLength,
+    initialPage,
```
ahmadtaimoor-deriv, line 16: So just to understand , the initialPage number will only be set on first render? and then on 
`onClickPagination` . We are not changing the initialPage from the consumer side ? 
If we are changing , I think we should use a better name for `initialPage`

lib/components/Pagination/index.tsx
```diff
@@ -30,8 +31,12 @@ export const Pagination = ({
...
 
     useEffect(() => {
-        onClickPagination({ currentPage, totalPageCount });
-    }, [currentPage, onClickPagination, totalPageCount]);
+        if (initialPage) setCurrentPage(initialPage);
+    }, [initialPage]);
+
+    useEffect(() => {
+        onClickPagination?.({ currentPage, totalPageCount });
+    }, [currentPage, totalPageCount]);
```
ahmadtaimoor-deriv, line 39: do we want to trigger onClickPagination with totalPageCount changes too?

kate-deriv: Good point, I don't think we should trigger. Will remove

lib/components/Pagination/index.tsx
kate-deriv: initialPage will be set not only  on first render, but in case if it changed too. Thre is a useEffect below for setting it into state

ahmadtaimoor-deriv (approved):

prince-deriv (approved):

PR #482, opened 2024-10-29
Akmal / fix: put snackbar behind modal

akmal-deriv:
This fix adjusts the z-index of the snackbar to ensure it appears behind modal dialogs, improving visual hierarchy and clarity.

1. Z-Index Adjustment: Modify the snackbar’s z-index so that it remains behind open modals, preventing visual overlap issues.
2. UI Consistency: Ensure that the snackbar does not obstruct modal content, enhancing focus on the modal dialog when active.
3. Improved User Experience: Maintain a clear and organized visual structure, allowing users to interact with modals without distraction.

This fix enhances the user interface by ensuring that the snackbar is appropriately layered behind modals.

### DIFF
--- a/lib/components/Snackbar/snackbar.scss
+++ b/lib/components/Snackbar/snackbar.scss
@@ -16,7 +16,7 @@
     background: var(--component-snackbar-bg);
     box-shadow: var(--core-elevation-shadow-330);
     position: fixed;
-    z-index: 999;
+    z-index: 149;
     bottom: -16px;
     transition: bottom var(--motion-duration-snappy)
         var(--motion-easing-inandout);


lib/components/Snackbar/snackbar.scss
```diff
@@ -16,7 +16,7 @@
     background: var(--component-snackbar-bg);
     box-shadow: var(--core-elevation-shadow-330);
     position: fixed;
-    z-index: 999;
+    z-index: 149;
```
kate-deriv, line 19: Should it be controlled? Maybe for soma cases we need max z-index

akmal-deriv: I don't really see anything that realistically uses any z-index between 149 and 999 so far, it's just the modal/portal that uses 150 and by this change I'm moving it behind it, so I'd say we don't need it now, and I'd add we shouldn't need this in the future if we properly set all z-indexes and not just 999, 9999, 9998, 99999, 100000, etc.

akmal-deriv: Unless, there are already cases where we want to show snackbar above modal, which is a pretty ridiculous thing to do since we should have consistency between z-hierarchy of these components, starting from design implementation

prince-deriv (approved):

PR #493, opened 2024-11-07
ci: added github-actions config to the dependabot configuration file

shayan-deriv:
## changes:
- added `github-actions` to the dependabot configuration


.github/dependabot.yml
shayan-deriv, line 1: the error reported : `Dependabot is not configured for Github-Actions`

ali-hosseini-deriv (approved):

PR #505, opened 2024-11-21
Suisin/chore: remove countries flag to reduce bundle size and fix blocked cards

suisin-deriv:
Previous size was 1.99MB
[Previous_Size.pdf](https://github.com/user-attachments/files/17839238/Previous_Size.pdf)

After removing country flags: 1.17MB
[Current_Size.pdf](https://github.com/user-attachments/files/17839246/Current_Size.pdf)


lib/components/Input/input-phone-number/country-flags.tsx
```diff
@@ -1,503 +1,505 @@
...
+    // FlagWallisAndFutunaIcon,
+    // FlagWesternSaharaIcon,
+    // FlagYemenIcon,
+    // FlagZambiaIcon,
+    // FlagZimbabweIcon,
 } from "@deriv/quill-icons/Flags";
 import { ReactNode } from "react";
 
 export const FlagsMap: Record<string, ReactNode> = {
     XX: <FlagPlaceholderIcon />,
```
amina-deriv, line 256: where are we using FlagPlaceholderIcon ?

suisin-deriv: FlagPlaceholderIcon is when the countryCode doesn't contain any flags then it will drop to default which is FlagPlaceHolderIcon 🙏

jia-deriv (approved):

lib/components/Input/input-phone-number/country-flags.tsx
```diff
@@ -1,503 +1,505 @@
...
+    // FlagVaticanCityIcon,
+    // FlagVenezuelaIcon,
+    // FlagVietnamIcon,
+    // FlagWallisAndFutunaIcon,
+    // FlagWesternSaharaIcon,
+    // FlagYemenIcon,
+    // FlagZambiaIcon,
+    // FlagZimbabweIcon,
 } from "@deriv/quill-icons/Flags";
 import { ReactNode } from "react";
```
amina-deriv, line 253: There is react import at the top as well

lib/components/Input/input-phone-number/country-flags.tsx
```diff
@@ -1,503 +1,505 @@
...
+    // FlagVanuatuIcon,
+    // FlagVaticanCityIcon,
+    // FlagVenezuelaIcon,
+    // FlagVietnamIcon,
+    // FlagWallisAndFutunaIcon,
+    // FlagWesternSaharaIcon,
+    // FlagYemenIcon,
+    // FlagZambiaIcon,
+    // FlagZimbabweIcon,
 } from "@deriv/quill-icons/Flags";
```
amina-deriv, line 252: for readablity, can we move import statements together and comment out all other lines
```
import React, { ReactNode } from "react";
import { FlagPlaceholderIcon } from "@deriv/quill-icons/Flags";

export const FlagsMap: Record<string, ReactNode> = {
    XX: <FlagPlaceholderIcon />,
};
```

amina-deriv (approved):

prince-deriv (approved):

PR #509, opened 2024-11-22
Suisin/chore: add scroll to selected items for InputPhoneNumber container

suisin-deriv:
https://github.com/user-attachments/assets/840d7535-f52c-4edc-a00a-6c6410d2f7ed

https://github.com/user-attachments/assets/8d302331-4a37-4813-b8c9-04a7e00abc9f

### DIFF
--- a/lib/components/Input/input-phone-number/dropdown-content.tsx
+++ b/lib/components/Input/input-phone-number/dropdown-content.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import { DropdownItem, ItemContainer } from "@components/Atom";
 import { useDropdown } from "@hooks/useDropdown";
 import useBreakpoints from "@hooks/useBreakpoints";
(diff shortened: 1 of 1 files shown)


lib/components/Input/input-phone-number/dropdown-content.tsx
```diff
@@ -44,24 +44,77 @@ const DropdownContent = ({
...
+    useEffect(() => {
+        if (isOpen && selectedItemRef.current) {
+            setTimeout(() => {
+                const selectedItem = selectedItemRef.current;
+
+                if (selectedItem) {
+                    if (isMobile) {
+                        // Get the search bar height
+                        const searchBar =
+                            document.querySelector(".phone-code-search");
```
prince-deriv, line 58: Can we have a more specific element targeting, ex: the parent container of the exact component. There is a low chance of having thesame class in another component but just to make sure that we are really getting the element from this component.

suisin-deriv: sure let me do that 🙏

prince-deriv (approved):

